PR #5, opened 2016-10-12
docs: RFC proposal to center talk about authentication around

bcoe:
I don't have particularly strong opinions about how authentication is implemented in yarn, but I've had a few customers ask after it already, and there are a few threads in the yarn issue tracker on the subject (https://github.com/yarnpkg/yarn/issues/521, https://github.com/yarnpkg/yarn/issues/547).

I thought it would be worthwhile to center this discussion around an RFC; it's a complicated enough feature that context could easily be lost across open GitHub issues.

### DIFF
--- a/text/0000-authentication.md
+++ b/text/0000-authentication.md
@@ -1,0 +1,119 @@
+- Start Date: 2016-10-11
+- RFC PR:
+- Yarn Issue:
+
+# Summary
+
+Add the ability to install from and authenticate against an alternative
+registry, via the environment variables `YARN_AUTH_TOKEN`, `YARN_REGISTRY`,
+and the `.yarnrc` settings `registry`, and `token`.
+
+# Motivation
+
+Services like npm, Artifactory, and Sinopia allow users to develop private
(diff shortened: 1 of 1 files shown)


joshmanders:
Yes! I would like to see something that supports context switching. I have a tool that I wrap around npm right now that allows me to login to multiple accounts and switch out the authentication token depending on a `--as` flag. Since I publish modules under my own account and employers account, sometimes it gets cumbersome logging out and into new accounts to make the changes, then back.

ljharb:
I do the same with an alias that moves around .npmrc files. It'd be great to have yarn more transparently use npm configs as opposed to inventing a non-drop-in way to configure things.

ashleygwilliams:
relevant: https://github.com/yarnpkg/yarn/pull/839

tl;dr this PR implements using a `.npmrc` file to auth to private registry

bcoe:
@ashleygwilliams yarnpkg/yarn#839 seems like a great approach to implementing authentication; Supporting multiple registries is definitely the best-case scenario.

bcoe:
It looks like @devongovett has pulled together a wonderful implementation of authentication/multiple-registry-support. Closing this RFC, and calling it a fun experiment 👍

PR #7, opened 2016-10-12
Create 0000-license-check.md

hzoo:
Original issue: https://github.com/yarnpkg/yarn/issues/904

Integrate a license checker as a yarn command.

And I have published a standalone package that does this (and am using): https://github.com/behance/license-to-fail

---

Example Usage

$ ./bin/license-to-fail ./config.js

Disallowed Licenses:
INDIRECT DEP - spdx-correct@1.0.2 Apache-2.0: https://github.com/kemitchell/spdx-correct.js

### DIFF
--- a/text/text/0000-license-check.md
+++ b/text/text/0000-license-check.md
@@ -16,7 +16,7 @@ Yarn has `yarn licenses ls`. It would also be useful to know if certain packages
 don't satify your license (or other similar files) requirements rather than just a list of them.
 
 ```bash
-$ yarn licesnes check
+$ yarn licenses check
 yarn licenses v0.14.0
 Disallowed Licenses
 ├─ a-pkg@1.0.3
--- a/text/text/0000-license-check.md
+++ b/text/text/0000-license-check.md
@@ -1,0 +1,100 @@
+- Start Date: 2016-10-12
+- RFC PR: 
+- Yarn Issue: 
+
+# Summary
+
+Original issue: https://github.com/yarnpkg/yarn/issues/904
+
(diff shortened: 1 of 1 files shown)


text/text/0000-license-check.md
```diff
@@ -0,0 +1,100 @@
...
+
+# Alternatives
+
+Just use a separate package rather than making it built-in like https://github.com/behance/license-to-fail already is (and others).
+
+# Unresolved questions
+
+How do users specify the allowed licenses and exceptions (differences for apps/libraries)?
+
+- use package.json config
```
bestander: I would prefer this one per project

hzoo: Yeah right now you pass a config LOC so you can share across an org

bestander:
Let's go with it 👍 
Waiting for PR

hzoo:
🎉 

Wow just realized the pr made the file at text/text for some reason..

bestander:
no worries, just send a pr to move the file when you have a chance

Kurt-von-Laven:
@noahnu wrote [a community-contributed plugin](https://yarnpkg.com/features/plugins#contrib-plugins) that adds license checking support to Yarn. I don't know how people feel about having a Yarn plugin vs. a package, such as license-to-fail. We have been using the licenses plugin for a few months now, and it works well for us. I wonder if there are features that would be needed to consider this RFC implemented (or, more formally, rejected since the plugin is not part of core); all I can say is it meets our needs. Full disclosure: I am a contributor to said plugin.

Issue #25, opened 2016-11-11
Add symlink link: dependency type


mgcrea:
We've been using some kind of monorepo approach for our projects for quite some time, a bit like [lerna](https://github.com/lerna/lerna) but with a more private and nested approach (our packages aren't expected to be published for now) along some specific needs: we must to link some public dependencies (eg. mongoose) that must be the exact same instance accross our subpackages (otherwise you encounter a lot of exotic bugs, edge cases), we also link our devDeps (they are shared accross all our subpackages).

At first we used [linklocal](https://github.com/timoxley/linklocal) with npm@2, leveraging a custom use of the `file:` prefix (basically just symlinking them), but npm@3 broke a lot of things related to their handling (eg. [#10343](https://github.com/npm/npm/issues/10343)). We ended up moving to [ied](https://github.com/alexanderGugel/ied) where we implemented the `file:` prefix handling using simple symlinks, that tackled our need.

I would love to be able to switch theses project to yarn but I would need a way to create this links, I can see two ways to achieve that:
- Either add a new `link:` prefix that would just create symlinks and that's it
- Add an option flag `--link-file-dependencies` that would override default `file:` handling (copying) and use symlinks instead.

--- 
Related Issue: https://github.com/yarnpkg/yarn/issues/884
Related PR: https://github.com/yarnpkg/yarn/pull/1109

whitecolor:
I believe this would solve many current issues with local deps.

whitecolor:
@mgcrea 
Is yarn cache going to be some how involved in the work with `link:` (will it be writing something to cache)?

shouze:
Other dependency managers like [composer](https://getcomposer.org/) manage that kind of feature 
with repository lists:
```json
    "repositories": [
        {
            "type": "path",
            "url": "../somelib",
            "options": {
                "symlink": true
            }
        },
        {
            "type": "vcs",
            "url": "https://github.com/someuser/somelib"
        }
    ]
```
The repositories order is important of course in the resolution of the dependency.

This is a very flexible way to do library nurturing in our mono repo whilst one day make that lib released either in a public or private npm repo. This is our use case with php libs and it's very convenient.

mgcrea:
@whitecolor yarn cache would not be involved.

@shouze,

I'm not sure how such a cascading repository resolution would work, does it mean if the local file does not exists it does a clone on GitHub? I personally would not need such feature, as I just want it to be dumb and create a symlink (even if the target does not exist yet).

[...]

PR #34, opened 2016-12-04
:link dependencies

mgcrea:
Following #25, hope the formatting is good enough (was unsure if I had to hard word wrap or not, and if yes how to deal with md links).

### DIFF
--- a/text/0000-link-dependency-type.md
+++ b/text/0000-link-dependency-type.md
@@ -29,15 +29,12 @@ prefix handling using simple symlinks, that tackled our need.
 I would love to be able to switch theses project to yarn but I would need a way
 to create these links.
 
+npm is also considering to add the same `link:` specifier, see this [recent RFC](https://github.com/npm/npm/pull/15900).
+
 # Detailed design
 
-I can see two (non exclusive) ways to achieve that:
-
-- Either add a new `link:` prefix that would just create symlinks and that's it
+We Add a new `link:` specifier that would just create symlinks and that's it
   (regardless of destination's existence)
-
-- Add an option flag `--link-file-dependencies` that would override default
-  `file:` handling (copying) and use symlinks instead.
 
 I've already implemented the changes in yarn's [pr#1109](https://github.com/yarnpkg/yarn/pull/1109) and I'm
 currently maintaining a fork since my team already rely on this for several
--- a/text/0000-link-dependency-type.md
+++ b/text/0000-link-dependency-type.md
@@ -1,0 +1,67 @@
+- Start Date: 2016-10-12
+- RFC PR:
+- Yarn Issue:
+
+# Summary
+
+Add symlink `link:` dependency type to enable complex cross-project development
+workflows.
+
+# Motivation
+
+This RFC is a spinoff of yarn's [issue #884](https://github.com/yarnpkg/yarn/issues/884).
+
+We've been using some kind of monorepo approach for our projects for quite some
+time, a bit like [lerna](https://github.com/lerna/lerna) but with a more private
+and nested approach (our packages aren't expected to be published for now) along
+some specific needs: we must to link some public dependencies (eg. mongoose)
+that must be the exact same instance accross our subpackages (otherwise you
+encounter a lot of exotic bugs, edge cases), we also link our devDeps (they are
+shared accross all our subpackages).
+
+At first we used [linklocal](https://github.com/timoxley/linklocal) with npm@2,
+leveraging a custom use of the `file:` prefix (basically just symlinking them),
+but npm@3 broke a lot of things related to their handling (eg.
+[#10343](https://github.com/npm/npm/issues/10343)). We ended up moving to
+[ied](https://github.com/alexanderGugel/ied) where we implemented the `file:`
+prefix handling using simple symlinks, that tackled our need.
+
+I would love to be able to switch theses project to yarn but I would need a way
+to create these links.
+
+# Detailed design
+
+I can see two (non exclusive) ways to achieve that:
+
+- Either add a new `link:` prefix that would just create symlinks and that's it
+  (regardless of destination's existence)
+
+- Add an option flag `--link-file-dependencies` that would override default
+  `file:` handling (copying) and use symlinks instead.
+
+I've already implemented the changes in yarn's [pr#1109](https://github.com/yarnpkg/yarn/pull/1109) and I'm
+currently maintaining a fork since my team already rely on this for several
+projects.
+
+# How We Teach This
+
+I think `link:` is pretty explicit, an update to the docs/cli-help should be
+enough.
+
+# Drawbacks
+
+Not sure how exactly cross-platform symlinks are today. However it looks like
+[Microsoft might be catching up](https://blogs.windows.com/buildingapps/2016/12/02/symlinks-windows-10/) on this issue.
+
+# Alternatives
+
+Beside the two alternatives exposed above, I can't think of anything else for
+now.
+
+Drawback of not implementing this is that we restrict how creative developers
+can be with complex multi-packages workflows.
+
+# Unresolved questions
+
+Not sure how we should handle actually publishing packages with such
+dependencies, the existing behavior for `file:` types?


text/0000-link-dependency-type.md
```diff
@@ -0,0 +1,67 @@
...
+Beside the two alternatives exposed above, I can't think of anything else for
+now.
+
+Drawback of not implementing this is that we restrict how creative developers
+can be with complex multi-packages workflows.
+
+# Unresolved questions
+
+Not sure how we should handle actually publishing packages with such
+dependencies, the existing behavior for `file:` types?
```
Spy-Seth: I think it could be safe to say that the `link:` prefix could only be used on private package (`private:true` in package.json). It will avoid a lot of difficulties and weird decision to manage all the publish/install case.

Spy-Seth:
In your point of view, how dependencies of the linked package should be managed? 

- Is they should be installed in the "parent" repo, or be installed in the linked repository? 
- Should they be installed before the link or will be installed on the fly with the other dependencies?

thejameskyle:
Is this going to be a necessary feature if we add workspaces by merging Lerna into Yarn?

benlangfeld:
Maybe I'm just being dense, @thejameskyle, but Lerna appears to be concerned only with managing separate packages contained within a monorepo, and I don't see how it would help with the use case where a monorepo contains many modules which are dependencies of each other. Am I missing something?

thejameskyle:
Lerna very much is designed around having cross-dependencies. The `lerna bootstrap` command is for linking them all together.

benlangfeld:
Ah, I'd completely misunderstood how Lerna works. Thanks for the pointer @thejameskyle.

mgcrea:
Bumping this to see if there is a chance it might get reviewed/accepted or if I should just close this for low interest?

thejameskyle:
The highest priority for Yarn is working through issues in existing functionality. If you want RFCs to be accepted sooner you can come help us out over here: https://github.com/yarnpkg/yarn/issues

moimikey:
this... _needs_ to be merged in. it's so essential for mono repo approaches!

thejameskyle:
So we're actually finalizing a spec (which I'll open up an RFC for) for what "Yarn Workspaces" will build in monorepos support to Yarn.

Once we have workspaces I don't see much of a use case for this RFC. I'm going to close it with the promise that we'll have something better.

thejameskyle:
Actually we should add this, it has valid use cases even with workspaces and npm is going to add it: https://github.com/npm/npm/pull/15900

thejameskyle:
Note that we need to align on behavior with npm. cc @iarna

iarna:
This seems in alignment with what npm is proposing. If you all spot any differences please speak up. Our spec goes into rather more detail about how this feature would be expressed in npm, but I don't think any of that is relevant to interoperability between the two projects.

ljharb:
Wouldn't the two projects need to agree on what goes in `package.json`, since it could end up published to the ecosystem?

mgcrea:
Great to have this feature back into discussion! Turns out I just updated my fork yesterday to rebase the changes, here is an up-to-date commit: https://github.com/mgcrea/yarn/commit/69f02c0c9ee1d1874c9a8aa5ca21adfbf30bfcf6

It's lacking test but will gladly add them back once we reach a PR-ready stage.

text/0000-link-dependency-type.md
```diff
@@ -0,0 +1,67 @@
...
+to create these links.
+
+# Detailed design
+
+I can see two (non exclusive) ways to achieve that:
+
+- Either add a new `link:` prefix that would just create symlinks and that's it
+  (regardless of destination's existence)
+
+- Add an option flag `--link-file-dependencies` that would override default
```
thejameskyle: Can we remove this from the RFC? Just add `link:`

mgcrea: @thejameskyle done.

iarna:
@ljharb I think the only point of differences is treating existing local directory dependencies (`file:/path/to/directory/`) just like `link:` dependencies. That is, `file:/path/to/directory/` becomes an alias for `link:/path/to/directory/`. But other than that I _think_ our RFCs are already aligned. Do you see any other differences?

iarna:
The only caveat is that we're actually still debating even _using_ `link:`. How do you all feel about just having `file:` create links and not introducing a new specifier? I'd like there to be some wider discussion around this…

mgcrea:
I personally wouldn't mind with only `file:` doing symlinks, but I guess there might be other people relying on the current `file:` mechanism. Adding the explicit `link:` and deprecating `file:` first sounds a bit safer (with a warning), I'd still go for that.

iarna:
I've updated the [npm RFC](https://github.com/npm/npm/pull/15900) to reflect the npm cli team's current consensus on this, which is that this should be a change to `file:` behavior and not introduce a new specifier.

It's worth noting that neither of the versions of my proposal for npm has us keeping legacy `file:` specifier behavior. All I changed was ditching a _duplicate_ specifier named `link:` that otherwise acts identically.

From npm's perspective this is a breaking change, which is why this is scheduled for npm@5. It's my suspicion that this change won't actually break anyone, but as we're changing the contract we'll bump the major.

benlangfeld:
What is left to decide here in order for this RFC to be accepted?

bestander:
@mgcrea, do you want to update this RFC to be inline with npm@5?

bestander:
However maybe I don't have enough context here but it feels like a complication merging `file:` and `link:` into a single specifier.
They are doing quite different things

iarna:
Making `file:` specifiers create links is drawing a line in the sand and saying that `file:` specifiers as they were implemented in `npm@2` were a mistake. They're not content addressable. (A problem for shrinkwraps and locks.) There's no way to know if what's in your `node_modules` is the same was what the specifier points at. (A problem for npm's model, which never trusts that your
 `node_modules` matches its shrinkwrap. Yarn avoids by pretending it can't happen. Which is fine, so long as your users never use other tools or do weird things.) TLDR: `file:` specifiers that copy are a mess. 

I'm wrapping up spec tests for the npm implementation over at [spec-local-specifiers](https://github.com/npm/npm/blob/link-specifier/test/tap/spec-local-specifiers.js) which y'all are welcome to crib.

bestander:
Ok, I am in to follow npm here

benlangfeld:
@bestander / @thejameskyle What can I do to contribute to this RFC and the matching implementation being merged?

bestander:
@benlangfeld, thanks for offering help.
Just hang on a few days, @arcanis the new Yarn's core team member, is getting ramped up on this issue.
After this we will start resolving all those issues around managing multiple local packages.

bestander:
FYI, we are planning to have final look at this on Thursday this week.
Workspaces is getting in active phase and I think `link:` will be needed for that, we dragged a little bit until now to make sure that we don't have conflicting features in Yarn that do the same job.

benlangfeld:
Thanks @bestander. At this point, I'm willing to pay a yarn subscription if it gets this merged; anything so my team can keep their hair.

If there's anything that can be done to help move this forward, don't hesitate to delegate :)

arcanis:
Hi! We read the proposal together with @bestander, and really like it. We think the feature described in this RFC might prove quite valuable, partly because of the use cases it contributes to solve, and partly because it might be a solid base for implementing more complex features later.

@mgcrea I saw that you opened a [pull request](https://github.com/yarnpkg/yarn/pull/1109/) with a prototype implementation, do you still want to champion with feature? If so, I'd be happy to reopen it so that we can discuss it in more details. Let me know :)

mgcrea:
@arcanis I'd be happy to. I'm off work this week so I can have some spare time to work on it.

Would we directly go for the (major?) breaking change of the `file:` dependency type like npm?

In that case, I can rewrite the PR accordingly, might need some extra-eyes to make sure there is no dead-code floating around (used exclusively by the current `file:` handling).

bestander:
IMHO breking file: is unnecessary at this moment.
Happy to discuss it in a new RFC

mgcrea:
Ok, I've quickly rebased my previous non-breaking work to https://github.com/yarnpkg/yarn/pull/3359 we can discuss implementation details there.

Issue #13, opened 2016-10-25
Custom dependency types (or groups)


creynders:
Having a limited number of dependency types does not make sense in today's development process. If I look at a typical project of mine I have various stages and environments during development and deployment. Breaking this down into just `devDependencies` and `dependencies` seems very insufficient.

It would be awesome to be able to declare them as such, e.g.:

```
dependencies: {
  dev: {
    babel: 'v0.0.0'
  },
  uitest: {
    nightwatch: 'v0.0.0'
  },
  unittest: {
    mocha: 'v0.0.0'
  },
  core: {
    express: 'v0.0.0'
  }
}
```

(Nvm the API, it's all about the concept)

Just to be clear the keys `dev`, `uitest`, etc. are arbitrarily chosen as an example. What I'd want to propose is the possibility to use any word you want. And group them in any way you want.

Installation could be as easy as e.g. `yarn install --core --uitest --dev` i.e. just specifying which dependency types/groups/whatever you want to install. I.e. you'd combine the keys to specify which types you want installed. E.g. on your build server you could do `yarn install --core --dev`, on your testing server you'd do `yarn install --core --unittest --uitest` And so on.

And maybe you should be able to declare a `default` entry (i.e. what `dependencies` is in npm-land)

This would allow us to install only what's really necessary on each machine: local dev, QA, testing, staging, production, ...

creynders:
Originally posted at yarnpkg/yarn#1431

thejameskyle:
Ignoring some of the problems that would have to be resolved with various apis having different meanings with the existing dependency types. This would cause issues with your use case:

> This would allow us to install only what's really necessary on each machine: local dev, QA, testing, staging, production, ...

The problem is that dependencies almost never fall into just one of those categories. You can't just force them into buckets like that.

They also don't fall under just "Unit Testing" or "Core" unless you want to introduce dependencies of categories as well ("Unit Testing" would have to depend on "Core" in order to create the correct set of dependencies when installed). 

In general I don't think we need to get that religious about making sure that people install the exact minimum number of dependencies that they have declared for the environment they are in. We can make installs fast enough that it shouldn't matter beyond the first install (that's true for CI environments as well).

creynders:
Ah sorry, apparently my `install` example was vague. I meant you would be able to specidy multiple groups. I.e on my testing server I'd do `install --core --unittest --dev` and on my build env I'd only do `install --core --dev` etc. You'd mix and match.

[...]

PR #35, opened 2016-12-05
Custom dependency groups proposal

creynders:
>Allow declaring dependencies in custom named groups (i.e. besides dependencies, devDependencies, ...)

Originally and incorrectly proposed in #13

### DIFF
--- a/text/0000-custom-dependency-groups.md
+++ b/text/0000-custom-dependency-groups.md
@@ -1,0 +1,288 @@
+- Start Date: 2016-12-05
+- RFC PR: (leave this empty)
+- Yarn Issue: (leave this empty)
+
+# Summary
+
+> Allow declaring dependencies in custom named groups (i.e. besides `dependencies`, `devDependencies`, ...)
+
+Instead of restricting dependency declarations to the pre-existing groups it would be beneficial to allow developers to decide for themselves how they want to organize their declarations.
+
+E.g. this:
+
+```json
+{
+   "dependencies": {
+     "bluebird": "3.3.5",
+     "backbone": "1.3.3",
+     "handlebars": "3.0.2",
+     "lodash": "3.10.1"
+   },
+   "devDependencies": {
+     "babel-core": "6.7.4",
+     "eslint": "2.11.1",
+     "handlebars-loader": "1.0.2",
+     "less": "2.6.0",
+     "less-loader": "2.2.2",
+     "mocha": "2.4.5",
+     "must": "0.13.1",
+     "nightwatch": "0.8.18",
+     "proxyquire": "1.7.4",
+     "require-directory": "2.1.1",
+     "sinon": "1.17.4",
+     "webpack": "1.12.13",
+     "webpack-dev-server": "1.14.1"
+   } 
+}
+```
+
+Could be re-organized into something like:
+
+```json
+{
+   "dependencies": {
+     "app": {
+       "bluebird": "3.3.5",
+       "backbone": "1.3.3",
+       "handlebars": "3.0.2",
+       "lodash": "3.10.1"
+     },
+     "build": {
+       "babel-core": "6.7.4",
+       "handlebars-loader": "1.0.2",
+       "less": "2.6.0",
+       "less-loader": "2.2.2",
+       "webpack": "1.12.13"
+     },
+     "QA": {
+       "eslint": "2.11.1"
+     },
+     "unittest": {
+       "mocha": "2.4.5",
+       "must": "0.13.1",
+       "proxyquire": "1.7.4",
+       "require-directory": "2.1.1",
+       "sinon": "1.17.4"
+     },
+     "uitest": {
+       "nightwatch": "0.8.18"
+     },
+     "dev": {
+       "webpack-dev-server": "1.14.1"
+     }
+   } 
+}
+```
+
+**Where the various group names (`app`, `build`, `QA`, ...) are developer-defined, i.e. they could be any string value.**
+
+Installation of the packages could be as simple as passing multiple group names to the command-line:
+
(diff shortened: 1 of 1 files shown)


text/0000-custom-dependency-groups.md
```diff
@@ -0,0 +1,288 @@
...
+
+# How We Teach This
+
+Documentation and education wise this would be an addition, in the sense that all existing documentation would still be valid. 
+
+Anybody familiar with "dependencies" and "devDependencies" would pick this up pretty fast, I think.
+
+# Drawbacks
+
+The major drawback would be that `package.json` files utilizing the new dependencies declaration style would no longer be recognized as valid manifest files by the `npm` CLI. I.e. once you go down this road, you can't use npm for dependency management in your project anymore.
```
thejameskyle: > once you go down this road, you can't use npm for dependency management in your project anymore.

Nor can any package consuming your package, you'd be opt-ing into an entirely new ecosystem.

text/0000-custom-dependency-groups.md
```diff
@@ -0,0 +1,288 @@
...
+
```
thejameskyle: Additional drawbacks:

- You never know what dependencies are actually needed for what. If you run `./scripts/do-something.js` while inside the repo, how do you know if you have the correct dependencies installed. You don't.
- You would still have to resolve the entire tree from the registry in order to be deterministic, this doesn't improve that at all.
- You complicate the install process so that every single project is different. You move from `git clone repo && yarn install` to `git clone repo && less package.json && yarn install --some --groups --that --hopefully --you --want`. Try something, doesn't work? Try again with the dependencies to install.
- You can't make it so that one category depends on another. ex. "test" needs "app", but there's nothing to say that)
- Dependencies don't fall neatly into any one category. ex. How do you say something is a "test" dependency vs a "build" dependency if it's being used for both? The line is never clear.

thejameskyle: Another one:

- It's ambiguous which dependencies are needed when installing a package as a dependency which uses this system. e.x. Do we install everything named "app" by default? What about the other categories?

^ This point actually makes this proposal impossible to implement

thejameskyle:
@wycats pointed out to me that Rust has something similar to this called [features](http://doc.crates.io/manifest.html#the-features-section).

The difference is that these all the equivalent of a specialized `optionalDependencies` where this proposal suggests making every dependency fall under a category.

text/0000-custom-dependency-groups.md
creynders: @thejameskyle thanks for your feedback! Regarding your last point, that's why I proposed a `default` group.
I'll make sure to add the drawbacks you listed. And clarify on installing `default` dependencies when installing a package as dependency.

creynders:
@thejameskyle I'll have a look at the rust features. I definitely believe there's value to this concept, but probably my implementation proposal should be improved to avoid (most of) the drawbacks you mention.
Thanks again for taking your time to comment, I really appreciate it.

text/0000-custom-dependency-groups.md
creynders: @thejameskyle So, I took some time to think about your comments. It seems to me you're not being entirely fair/honest about some of the drawbacks, in the sense that some of them aren't specific/unique to my proposal at all:

> If you run ./scripts/do-something.js while inside the repo, how do you know if you have the correct dependencies installed. You don't.

Surely this is always a problem? If you installed the production dependencies but didn't install the devDependencies for instance. Granted, with dependency groups the effect is enhanced.

> You would still have to resolve the entire tree from the registry in order to be deterministic, this doesn't improve that at all.

True, but you'd still have a speed improvement since not all the packages get installed, no? Or is the bulk of the time cost spend on tree resolution?

> You complicate the install process so that every single project is different. You move from git clone repo && yarn install

I don't see how it's any different in my proposal? Most projects make sure `npm install` or `yarn install` is what gets you up and running, but then provide a guide on what to do in case you want to contribute and/or test, which is almost never a simple `yarn install` but rather involves some configuration and setting up. Adding a few switches to this set up process doesn't seem to be such an extra complication to me.

> Dependencies don't fall neatly into any one category. ex. How do you say something is a "test" dependency vs a "build" dependency if it's being used for both? The line is never clear.

Sure. But it's up to you to decide what makes sense to you. There's so many things in development and devops where the line isn't clear.

doug-wade:
A similar idea, that might be more compatible with the existing npm infrastructure, might be to support "features" or "custom named groups" as a separate member of package.json, e.g.

```
{
  "dependencies": {
    "chalk": "*"
  },
  "devDependencies": {
    "ava": "*"
  },
  "features": {
    "test": ["ava", "chalk"]
  }
}
```

and then support those features in _prune_ rather than install.  Then if you wanted to have a working module, you run a `yarn install`, but if you only want the "test" feature dependencies installed, you can run a `yarn install && yarn prune --test` to prune down to only the dependencies you want.

This would allow packages to stay in the existing ecosystem, since the normal dependency hashes (dependencies, devDependencies &c) to stay intact, would allow dependencies to be part of multiple features (the packages that remain in node_modules are the union of all features), could be extensible and would allow devs who want this module to "just work" :tm: to run `yarn install` as usual without guessing, but would still allow me to catch devDependencies that should be listed as dependencies on CI.

thejameskyle:
@doug-wade that's a much more feasible proposal, would have to think it through more.

creynders:
@doug-wade I agree with @thejameskyle that sounds as a really good idea. Will you set up a separate RFC?

doug-wade:
@thejameskyle @creynders I'll totes write something up this evening

creynders:
Closing this in favor of #36

PR #36, opened 2016-12-08
add initial draft of feature pruning

doug-wade:
Bring [Rust's features](http://doc.crates.io/manifest.html#the-features-section) to yarn 🎉 🎆

### DIFF
--- a/text/0000-feature-pruning.md
+++ b/text/0000-feature-pruning.md
@@ -222,3 +222,6 @@ Should `yarn prune` without arguments or flags continue to recommend
 
 Should regular dependencies always be retained, even if the `--prod` flag is not
 provided?
+
+Should we add an `--exclude` switch for install, which would install all 
+dependencies except for the ones declared in the named feature?
--- a/text/0000-feature-pruning.md
+++ b/text/0000-feature-pruning.md
@@ -68,7 +68,7 @@ correspond to flags passed to `yarn prune`.  The values associated with those
 keys are always lists, consisting of either objects or strings.  If passed a
 string, it must be one of the following: a key in the `dependencies`,
 `devDependencies`, `peerDependencies` or `optionalDependencies`; the name of
-another key in the `features` hash; or one of `dependencies`,
+another key in the `features` hash; or one of the strings `dependencies`,
 `devDependencies`, `peerDependencies` or `optionalDependencies`.  If the
 name of a dependency is provided, that dependency is a member of the feature; if
 the name of a feature is provided, all dependencies in that feature are included
--- a/text/0000-feature-pruning.md
+++ b/text/0000-feature-pruning.md
@@ -1,0 +1,224 @@
+- Start Date: 2016-12-07
+- RFC PR: (leave this empty)
+- Yarn Issue: (leave this empty)
+
+# Summary
+
+> Allow organizing dependencies in custom named groups
+
+At different points in a package's lifecycle, the package may require a
+different set of dependencies, currently represented as `devDependencies`,
+`peerDependencies`, `dependencies` and `optionalDependencies`.  While these sets
+of dependencies are expressive for the majority of existing node packages
+published on the public registry, some packages, and many applications, have
+lifecycles that are much more complicated than than the develop and deploy
+lifecycle that currently exists.  This rfc allows for packages to define custom
+dependency closures to match their lifecycle.
+
+# Motivation
+
+Better support for continous integration pipelines. Currently, there is a class
+of bugs that is possible to write that cannot be caught with npm test. Consider
+the following: you have a node module for which you've dutifully written unit
+tests and listed your test framework as a dev dependency so that contributors
+continuous integration server can build and test your package with the same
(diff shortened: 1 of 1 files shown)


doug-wade:
You may also want to check out some prior art: https://github.com/yarnpkg/rfcs/pull/35 and https://github.com/npm/npm/issues/11899.

creynders:
What do you think of an `--exclude` switch for `yarn install`? Sth like:

```sh
$ yarn install --exclude test
```

Which would install all dependencies except for the ones declared in the `test` feature.

text/0000-feature-pruning.md
```diff
@@ -0,0 +1,224 @@
...
+`peerDependencies`, `dependencies` and `optionalDependencies`.  While these sets
+of dependencies are expressive for the majority of existing node packages
+published on the public registry, some packages, and many applications, have
+lifecycles that are much more complicated than than the develop and deploy
+lifecycle that currently exists.  This rfc allows for packages to define custom
+dependency closures to match their lifecycle.
+
+# Motivation
+
+Better support for continous integration pipelines. Currently, there is a class
```
doug-wade: Amusingly, I made this mistake just today -- I used [np](https://github.com/sindresorhus/np) to publish 0.0.2 of [react-link-to-inbox](https://www.npmjs.com/package/react-link-to-inbox) and listed [styled-components](https://github.com/styled-components/styled-components) as a devDependency instead of a regular dependency :joy:

doug-wade:
@creynders the flag should be `--include` rather than `--exclude`, since you want to _include_ the dependencies in the feature that you've listed, rather than exclude them, right?

creynders:
@doug-wade Erm... I thought the proposed functionality is to have all dependencies installed automatically? Then it would be useful to be able to tell _not_ to install specific features.

text/0000-feature-pruning.md
```diff
@@ -0,0 +1,224 @@
...
+
+I also propose adding a new cli command, `yarn prune`, named for the equivalent
+[npm command](https://docs.npmjs.com/cli/prune), which it approximates.  For npm
+compatibility, a `yarn prune`, with no arguments, would remove any dependencies
+that are not needed.  In this way, it is similar to a `yarn install --force`,
+except that it is strictly subtractive, rather than additive.  
+
+However, unlike the existing npm command, when flags are passed to `yarn prune`,
+the set of dependencies specified by the feature whose name corresponds to the
+flag will be retained, as well as all transitive dependencies, while the rest of
```
creynders: I understand why you propose it like this, but it seems weird that the passed argument is retained, _not_ pruned. I.e. intuitively I'd think that `yarn prune --build` will prune the "build" feature, not retain it.

doug-wade: yup, but I think it would be even weirder if `--production` had the opposite effect of every other argument.

doug-wade:
@creynders I added the switch as an open question.

doug-wade:
I would really appreciate some feedback from the yarn team on this one.  I understand y'all have been very busy, but it's been 2 months without any acknowledgement or guidance -- I would really appreciate even a cursory review.  @bestander do you have a few minutes to read over this?

bestander:
I am looking at this from a position of support.

So in the motivation, would the whole problem be resolved by having an e2e test with `yarn install --production`?
The advantage of this feature is saved disk space?
Because if you would have to run --prune after installation then it would not save installation time.
I wonder if this feature could be implemented as a utility if Yarn would provide a JS API to resolution and hoisting.

doug-wade:
@bestander thank you so much for taking the time to read this over.  I understand your concerns about support, and I'm very empathetic.  To answer your questions:

> would the whole problem be resolved by having an e2e test with yarn install --production?

Possibly, but then how do you install the test utilities?  If you have control over your CI host, like a Jenkins box, then you can globally install it, but now every time you change your e2e testing tool version, you have to install it on every dev box and every CI box; if you're using something like travis + appveyor, you have to track your e2e testing tool in three places: travis.yml, appveyor.yml, and CONTRIBUTING.md.  It might work, but features offers a much more elegant solution.

> The advantage of this feature is saved disk space?

As you point out, you don't save disk space or installation time.  The motivation is _correctness_, for instance, in the case that I [mentioned earlier](https://github.com/yarnpkg/rfcs/pull/36#discussion_r91616793).  The dependency, styled-components, was available locally because I installed it as a devDependency, so my tests passed.  However, the first person who used the new version of the package (luckily it was me) was greeted with a dependency missing error.  With features, I could list my e2e testing tool in a `--testing` feature, and then run `yarn prune --production --testing` and then run my tests, and I would get the missing dependency error before the PR is merged.

bestander:
Not 100% sure it will be an improvement for the dev experience.

Why would I want to prune dependencies in the first place?
Why would not I just install all node_modules for all the use cases for my app?

I think splitting the app into multiple sub apps with their own package.json files makes more sense.

doug-wade:
I'm clearly swimming upstream here.  Ima close this one.

PR #38, opened 2016-12-26
auto-build-rfc

hemanth:
//cc @samccone @kittens 

#11

### DIFF
--- a/text/0000-auto-build.md
+++ b/text/0000-auto-build.md
@@ -1,0 +1,59 @@
+- Start Date: 2106-12-26
+- RFC PR: (leave this empty)
+- Yarn Issue: (leave this empty)
+
+# Summary
+
+Auto build on npm link/post install.
+
+# Motivation
+
+`git clone` yarn repo and `npm link` results in:
+
+``` sh
+$ yarn install
+module.js:339
+    throw err;
+    ^
+
+Error: Cannot find module '../lib-legacy/constants'
+    at Function.Module._resolveFilename (module.js:337:15)
+    at Function.Module._load (module.js:287:25)
+    at Module.require (module.js:366:17)
+    at require (module.js:385:17)
+    at Object.<anonymous> (/Users/hemanth/labs/yarn/bin/yarn.js:9:17)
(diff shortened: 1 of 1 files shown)


ljharb:
@hemanth if you want a postinstall script, couldn't you add one under "postinstall"?

hemanth:
Yes, that's what I meant in [yarn#1416](https://github.com/yarnpkg/yarn/issues/1416).

Daniel15:
Do any other packages automatically build when you run `npm install`? That feels a bit too magical for me 😛 

What is the purpose of this change?

hemanth:
For making things easier with `git clone` and `npm link`.

Daniel15:
Sure, but what's the end goal? 😃 

For a user of Yarn that wants to use the latest version, we have nightly builds available to use (https://yarnpkg.com/en/docs/nightly)
For someone building on Yarn itself, you'd need to know how to build Yarn, so I'm not sure that automating just the first build will actually help, unless people end up understanding exactly why it happens. 

I'm not against the change at all, just curious as to its motivation.

hemanth:
> I'm not against the change at all, just curious as to its motivation.

Laziness being the main motivation ;) 

On a serious not, say someone is making a PR to yarn repo, if `npm link` does the required, it would be one step lesser and easier?

Daniel15:
> say someone is making a PR to yarn repo, if npm link does the required, it would be one step lesser and easier?

For what it's worth, I don't use `npm link` when building on Yarn. I just run it with the exact absolute path (for example, `c:\src\yarn\bin\yarn` or `/usr/local/src/yarn/bin/yarn`). It makes it immediately obvious exactly *which* version of Yarn I'm running :)

doug-wade:
@hemanth why aren't we proposing use of the [npm `prepare` script](https://github.com/npm/npm/releases/tag/v4.0.0) for this?  It seems to me that following the npm pattern would be better for interoperability, and make this easier.

bestander:
done via https://github.com/yarnpkg/yarn/pull/3553

PR #40, opened 2017-01-10
Initial RFC for publishConfig for yarn

DarthHater:
This is an initial draft for adding the ability to use publishConfig with yarn.

@bestander this is functionality that I think would be very useful for use with Nexus Repository Manager, Artifactory, etc...

### DIFF
--- a/text/0000-publish-config.md
+++ b/text/0000-publish-config.md
@@ -1,0 +1,53 @@
+- Start Date: (1-10-17)
+- RFC PR: (leave this empty)
+- Yarn Issue: (leave this empty)
+
+# Summary
+
+Right now yarn does not have a concept of publishConfig. This setting already
+exists in package.json for many npm packages. The setting allows you to set the
+registry url where you want your package published.
+
+# Motivation
+
+The addition of picking up publishConfig from package.json will allow developers
+creating internal packages to move across to using yarn with greater ease.
+
+Developers will be able to use yarn publish to send their package to an internal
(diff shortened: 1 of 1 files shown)


ljharb:
+1, this would be great. publishConfig is an important safety measure that npm respects, and it would be potentially very disastrous to run `yarn publish` in a previously npm-managed package, and accidentally make a private package irrevocably public.

DarthHater:
@ljharb I think there are currently measures to prevent that, yarn seems to pick up the private variable, so those would get shortcut, IF you have that set.

ljharb:
Correct, but that's a big "if". Often publishConfig is the only mechanism - and that was sufficient with npm.

bestander:
Considering that this feature is already in npm I agree, this should be implemented in Yarn.
FYI, we are planning to work big on publish command in the next 3-4 months to bring some [lerna](https://github.com/lerna/lerna) features

DarthHater:
@bestander this would be cool to add in the short term, I think it would be very useful for users using a private registry. I've got a branch where I'm working on the implementation, pending approval over here.

Daniel15:
Nice! I didn't know about publishConfig. I agree with this 😃 

This is somewhat related to https://github.com/yarnpkg/yarn/issues/547, where I mentioned that we should have better handling for custom npm servers.

bestander:
@DarthHater, looks like no one objects and this seems like a very important feature.
Please send the PR when you have a chance :)

PR #42, opened 2017-01-15
Global Flat Linked Package Folder with no bloating node_modules in each project 

vairamsvsjdo:
Global Flat Linked Package Folder with no bloating node_modules in each project.

### DIFF
--- a/0000-default-global-symlink-flat-packages.md
+++ b/0000-default-global-symlink-flat-packages.md
@@ -20,19 +20,19 @@ Also, there is not an efficient way of avoiding node_modules tree hierarchy hell
 # Detailed design
 
 ## Avoid Name collision using namespaces in Semver Extend packages and store as tar balls (gz, zip) all packages using the following format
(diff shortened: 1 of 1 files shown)


bestander:
This is how Yarn was intended to work originally but this did not work for Node.js because too many modules depend on the physical structure of folders.
For more info see this discussion https://github.com/yarnpkg/yarn/blob/master/__tests__/commands/install/integration.js#L306

bestander:
At this stage - we are not ready to get this feature in.
As discussed quite a few times there are many CONs to this approach.
But I would suggest this - let's make the `linking` phase of installation, when files are copied from cache to node_modules, pluggable.
So that users could implement their own strategies for linking, then global symlinking could be one of them.

vairamsvsjdo:
What are other benefits of yarn over npm?

vairamsvsjdo:
How does yarn will achieve caching if dependencies are nested.
Will it download dependencies twice or it copies flat and adds to the nest?

pauldraper:
See #2125

PR #43, opened 2017-01-16
Torrent dependencies

emschwartz:


Daniel15:
Interesting idea! I think we still use BitTorrent internally at Facebook for wide-scale deployment (https://research.fb.com/publications/development-and-deployment-at-facebook/, http://arstechnica.com/business/2012/04/exclusive-a-behind-the-scenes-look-at-facebook-release-engineering/). 

My understanding is that BitTorrent works best with a small number of torrents containing a small number of large files. I'm not sure how well it'd work for npm dependencies, as there's generally a large number of dependencies which generally contain a large number of small files. Additionally, using torrents requires ports on the firewall to be opened.

I feel like it's an interesting idea, but torrents wouldn't really have many advantages due to the small file size of most npm packages. I feel like it'd be easier to simply specify HTTP mirrors, and you still get some of the same advantages (like downloads being distributed across multiple endpoints).

deathgaze:
I like this idea, but it's questionable as to whether it would work behind corporate firewalls/proxies.

emschwartz:
Fair points about file size and firewalls.

Regarding file size, I wonder if having many peers -- possibly closer than the remote registry and more than likely on the local network -- would still make it more efficient. A big part of the draw of yarn for me is not having to re-download the same dependency if I already have it installed on my machine. It seems like a natural next step to avoid downloading it from the registry again if a copy already exists on a machine in my local network.

@deathgaze out of curiosity, does the demo on https://webtorrent.io work for you at work (it uses WebRTC)?

bestander:
Nice idea!
My gut feeling is that this feature won't be used a lot so it would be hard to keep it maintained and justify spending time on maintaining it.

However that could be a great community project that could plug into Yarn.

What if Yarn could expose more JS API for plugins?
Something like
```
<install yarn>
yarn add-plugin torrent-resolver
yarn add torrent:my-package@version
```

victornoel:
Hi,

Some ideas on the matter: there is existing package manager that relies on p2p for distributing packages, see for example [pacserve](http://xyne.archlinux.ca/projects/pacserve/) for Archlinux (basically, cache directories are shared across a LAN so that users can directly download dependencies from a peer instead of from the original repository) and [apt-p2p](http://www.camrdale.org/apt-p2p/) for Debian (more internet-wide system, not really meant to be used internally of a LAN but it can).

Note that both of these do not use torrents, simply because it is not really adequate for this, it would almost be better to implement your own protocol on top of DHT (over the internet or inside a LAN) or even multicast discovery (inside a LAN).

Also I personally feel that this kind of feature does not really make sense at the scale of the planet. The real use case for using p2p sharing of packages is inside a LAN. To me pacserve is really a great example of a working system: your computer will download from peers if it can, packages are signed so you don't download something strange and the package manager falls back to typical repositories/registries when no peer has the desired package. It speeds things up a LOT without the need to introduce a registry cache internally.

Hope it's useful :)

emschwartz:
Thanks @victornoel for pointing that out. 

I agree that the main benefit would be downloading packages from peers on the LAN, but I think that would be huge for Node.js developers. At least for me, the normal setup is me sitting in an office with people working on the same set of repos using the same set of dependencies. I think it would make a lot of sense to be able to fetch dependencies from my coworkers.

Whether it would make more sense to use torrents or implement something custom, I'm not sure. In general I'd prefer to reuse work that's already been done.

victornoel:
@emschwartz yes, the use case you describe is exactly the one I think is a real use case.

Regarding implementation, what I meant is that torrent is clearly NOT adequate for this.
I don't know if there exists some libraries or protocol for sharing packages (which is a quite particular use case, very different from the torrent use case one), but for sure going towards the torrent direction is bound to fail IMHO.

Still, note that there exists a torrent based debian package sharing system: http://debtorrent.alioth.debian.org/
I think it is dead though.

cpojer:
I don't think at this point we'll be moving forward with this proposal. However, in the future we may make the API more extensible which could allow for support of this feature on top of Yarn.

PR #45, opened 2017-01-31
Support per-dependency target environments & record them in `package.json` file & `yarn` cli;

hulkish:


hulkish:
https://github.com/yarnpkg/rfcs/issues/46

hulkish:
?

doug-wade:
@hulkish this breaks interop with npm, and is unlikely to get merged.  Maybe #36 would work for your use case instead?

bestander:
What are you trying to achieve?
Speed of installation?
Categorization to make package.json more readable for developers?

Would be nice to see some real world example with the problem and how it is being managed right now

hulkish:
@bestander Well,  tooling aside I think it'd be helpful to know the nature of use for each included dependency. Simply "dev"  "peer" or "normal"  dependencies dont account for this.

bestander:
I don't really have a strong opinion on this.
It sounds like it can be convenient to categorize large dependency lists but it does not seem to be helping much for very large package.json files anyway.

hulkish:
close this... i guess?

PR #51, opened 2017-03-07
Prevent Yarn from rewriting the lockfile when switching between online ⇄ offline 

arcanis:
First draft cc @bestander


cpojer:
This is great!

0000-offline-resolution-field.md
```diff
@@ -0,0 +1,73 @@
...
+
+When enabling the offline mirror, Yarn updates the lockfile by stripping the registry URL from its `resolved` fields. This RFC aims to simplify this process by making such an update unneeded.
+
+Related issues: https://github.com/yarnpkg/yarn/issues/393 / https://github.com/yarnpkg/yarn/issues/394
+
+# Motivation
+
+Yarn currently has two different type of values for the lockfile `resolved` fields:
+
+  - When online, they're in the form `${source}/${name}-${version}.tar.gz#${hash}`
```
bestander: also it can be a git URL e.g. ssh://git@bitbucket.org/bestander/test-repo.git, a link: or a file:.

0000-offline-resolution-field.md
```diff
@@ -0,0 +1,73 @@
...
+
+# Detailed design
+
+I suggest the following:
+
+  - Removing the `${source}` part of the `resolved` field
+
+    - **Rational: Why removing the source part of the resolved field?**
+  
+      The source is redundant, and misleading:
```
bestander: Another advantage is that Yarn can start downloading the dependencies immediately skipping the resolution step.
Also there may be some caveats regarding proxies and private packages.

bestander:
I wonder if getting rid of domain from URL and making yarn.lock not changed when using Offline Mirror are two different problems you are trying to tackle.
Would it be simpler to do deal with them separately?

0000-offline-resolution-field.md
arcanis: > Also there may be some caveats regarding proxies and private packages.

It might also causes issue on this regard even with the source; for example, a dev might use an internal IP as mirror, and a public URL on CI. Baking the source in the lockfile means that this cannot be done without postprocessing the lockfile (since the CI won't have access to this URL). Removing the source would help resolve this issue, since then the CI would use whatever registry it is configured to use rather than the ones specified in the lockfile.

arcanis:
A first iteration could uniformize the two processes by baking the source into both offline & online package resolutions. There might be a bit less code involved, but I'm not completely sure (because there wouldn't be any distinction between offline and online, we would have to check the offline cache anyway to see if a matching file exists). The only step that would be avoided would be to dynamically resolve the source against the package.json (because then we would just use whatever source is set in the lockfile instead).

bestander:
Can you give an example how yarn.lock will look?

arcanis:
If we strip the source: (note that even the git repo is only resolved to its name + hash)

```
"@manaflair/term-strings@https://github.com/manaflair/term-strings.git#0782bd21a953f29f6425412865b826a62a06722c":
  version "0.7.0"
  resolved "term-strings-0782bd21a953f29f6425412865b826a62a06722c.tar.gz#0782bd21a953f29f6425412865b826a62a06722c"
  dependencies:
    babel-cli "^6.18.0"
    color-diff "^1.0.0"
    core-js "^2.4.1"

abbrev@1:
  version "1.1.0"
  resolved "abbrev-1.1.0-d0554c2256636e2f56e7c2e5ad183f859428d81f.tgz#d0554c2256636e2f56e7c2e5ad183f859428d81f"

ajv@^4.9.1:
  version "4.11.4"
  resolved "ajv-4.11.4-ebf3a55d4b132ea60ff5847ae85d2ef069960b45.tgz#ebf3a55d4b132ea60ff5847ae85d2ef069960b45"
  dependencies:
    co "^4.6.0"
    json-stable-stringify "^1.0.1"
```

If we forcefully write the source even when offline:

```
"@manaflair/term-strings@https://github.com/manaflair/term-strings.git#0782bd21a953f29f6425412865b826a62a06722c":
  version "0.7.0"
  resolved "https://github.com/manaflair/term-strings.git#0782bd21a953f29f6425412865b826a62a06722c"
  dependencies:
    babel-cli "^6.18.0"
    color-diff "^1.0.0"
    core-js "^2.4.1"

abbrev@1:
  version "1.1.0"

ajv@^4.9.1:
  version "4.11.4"
  resolved "https://registry.yarnpkg.com/ajv/-/ajv-4.11.4.tgz#ebf3a55d4b132ea60ff5847ae85d2ef069960b45"
  dependencies:
    co "^4.6.0"
    json-stable-stringify "^1.0.1"
```

0000-offline-resolution-field.md
bestander: Agree but it is a separate issue/feature

0000-offline-resolution-field.md
```diff
@@ -0,0 +1,73 @@
...
+      
+        - If we change our registry URL, the package itself will not have changed. The lockfile should not be modified.
+        
+        - The source is already in the package.json file - we can use it when we need to, since we will have checked by then that the resolved version still match (when working with a registry). Worst case scenario (direct link) we will catch that the file will have changed when checking the hash - but that's not something we could have avoided with storing the source in the lockfile anyway.
```
bestander: The source at package.json is loose that is resolved to a fixed URL in package-resolver.
Yarn gains speed by caching and freezing resolution step in the lockfile

0000-offline-resolution-field.md
```diff
@@ -0,0 +1,73 @@
...
+
+  - (Optional) Splitting the resulting `resolved` field into two others: `hash` (the hash of the tarballs) and `tarball` (the tarball name).
```
bestander: That might be useful especially that we want to migrate to sha2/sha3

0000-offline-resolution-field.md
```diff
@@ -0,0 +1,73 @@
...
+
+  - Adding the hash to the tarball name. This is to prevent issues where two projects are being worked on at the same time, both with the same dependency `foobar`, except that one of those projects is using the development branch of `foobar` (possibly fetching the project directly from git). In such a case, both projects lockfile would resolve to the same dependency name, the same dependency version (because maintainers don't usually bump version number until the release), but different hashes.
```
bestander: Nice idea

arcanis:
I opened a PR for review (https://github.com/yarnpkg/yarn/pull/2970) and updated the rfc (fortunately, there was more to remove than to add 😄).

bestander:
This looks good.
Can you address git dependencies in the RFC?
There we did not just strip source we actually generated a whole new file name some times.

bestander:
Great job, one big source of inconsistent caches is about to go away!

PR #53, opened 2017-03-14
Add the option to not use a cache

dguo:
Here is a [rendered version](https://github.com/dguo/rfcs/blob/no-cache/text/0000-no-cache-option.md).

This RFC is related to yarnpkg/yarn#986.

### DIFF
--- a/text/0000-no-cache-option.md
+++ b/text/0000-no-cache-option.md
@@ -1,0 +1,62 @@
+- Start Date: 2017-03-14
+- RFC PR:
+- Yarn Issue: [yarnpkg/yarn#986](https://github.com/yarnpkg/yarn/issues/986)
+
+# Summary
+
+Allow users to bypass the global cache.
+
+# Motivation
+
+For certain situations, it would be preferrable to not use a global cache for
+packages because the cache takes space and the step of copying from the cache
+to `node_modules` can take a significant amount of time. Some examples are:
+
+1. Building a Docker image. Images are isolated, so there's no point in having
+a cache within the image. If anything, it has the negative effect of increasing
+the image size.
+
+2. Running Yarn in a CI context. This is another example of a potentially
+isolated environment where we would just want to put the installed packages
+directly into `node_modules`. While sophisticated CI setups might have
+multiple projects share the same cache, there are certainly setups that don't
+do that.
+
+3. Running Yarn on a long-running server. The cache grows indefinitely over
+time, eventually reaching max disk size or inode limit. Simply bypassing the
+cache at run time could be preferrable to having to periodically run `yarn cache
+clean`.
+
+Overall, the desired effects of bypassing the cache are to increase the speed
+and decrease the disk usage of certain Yarn commands (i.e., install, add,
+upgrade).
+
+# Detailed design
+
+There should be a CLI switch `--no-cache` that tells Yarn to skip doing any work
+with the cache.
+
+Furthermore, there could be an equivalent `.yarnrc` setting of `no-cache` that
+does the same thing when set to `true`. This way, the developer doesn't need to
+remember to use the CLI switch every time.
+
+# How We Teach This
+
+We would update the [documentation](https://yarnpkg.com/en/docs/cli/cache) for
+the `yarn cache` command.
+
+# Drawbacks
+
+Increases the complexity of the implementation of Yarn.
+
+# Alternatives
+
+In the linked issue, there is a suggestion to set the cache folder to some
+in-memory or temporary location, but this solution is hacky and doesn't really
+help with the speed concern because Yarn still has to do the work of copying
+the packages to `node_modules`.
+
+The same reasoning applies to the option of immediately running
+`yarn cache clean`.
+
+# Unresolved questions


felixfbecker:
Use case 3:
Running yarn as part of a long-running server. The cache grows indefinitely over time, eventually reaching max disk size or inode limit.

dguo:
@felixfbecker, I added it to the doc.

felixfbecker:
Thanks! I would like to point out that `yarn cache clean` (besides being slower than just disabling it) is not possible if there are `yarn install`s running concurrently, because they all share the same global cache.

bestander:
Indeed cache introduces a lot of issues and it would be nice to have an option avoid it.
Although I find it hard to implement.

Here is a use case: babel-runtime can be used transitively 5 times in one project.
With cache it will be downloaded once from the registry, unzipped and then during linking phase it would be copied 5 times into node_modules.

If you try to avoid cache then would it download and unzip the package 5 times?
Even if we are ok with this it would complicate the phases in Yarn: resolver, fetcher and linker would have to interface with each other.

felixfbecker:
@bestander If babel-runtime is depended on 5 times _on the same version_, the package should not have to be duplicated, it would be flat in node_modules (what you are describing sounds like old NPM v2 style). If the versions are different, you need to download a different archive anyway.

ljharb:
node_modules is never guaranteed to be flat, it's just (in npm 3 and later) as flat as *possible*. The dependency graph might require the duplication.

felixfbecker:
@ljharb in what cases does it require duplication if the versions are the same?

ljharb:
@felixfbecker if A and B require X v1, and C and D require X v2, you are virtually guaranteed to get 2 copies of one version of X, and the other version would be deduced at the top level - totaling 3 copies.

felixfbecker:
True. yarn could be made smart enough in that case to copy Xv1 from A to B and Xv2 from C to D (or if the option is set, hardlink them).

bestander:
Yeah, from implementation point of view it might require too many code structure compromises.
I think we can achieve the goals some other ways.
I propose closing this one.

felixfbecker:
@bestander what would you propose?

bestander:
Each of those cases can be addressed with a workaround.

1. Building a Docker image => just use a local cache folder that you clean after install
2. Running Yarn in a CI context. => clear caches before run
3. Running Yarn on a long-running server. => clear caches

felixfbecker:
Unless I'm missing something, a long running server can't clean caches between runs, because many installations for different folders could run at any time in parallel. It doesn't seem like a good idea to clean the cache while a different installation is in the middle of the fetch or linking phase.

felixfbecker:
That is correct and the workaround we are currently using. But this is a performance-critical application (which is also why we use yarn, not npm), and the linking phase takes up a considerate amount of time. If I understand it correctly, that is the time spent copying/linking packages from the cache to node_modules?

bestander:
Yeah, linking phase is copying files.
Maybe using move instead of copy + --link-duplicates could be a less disruptive change.

felixfbecker:
Wouldn't move be subject to the same concurrency issues as cleaning the cache?

bestander:
If you use a unique cache folder for each install then no

felixfbecker:
Oh, right. I wonder if move is significantly faster than hardlinking though?

bestander:
If you move a folder it is super fast.
Moving files individually might not

felixfbecker:
Then adding an option to move would be great. It's unfortunate `--link-duplicates` can't be used for the time being because of https://github.com/yarnpkg/yarn/issues/2734

Diokuz:
Any chances to have an options, which disables cache? I have a docker, and I have to run `&& yarn clear cache` during build process.

bestander:
@Diokuz, it may be hard to implement at this moment.
I think it would be easier to introduce a robust hack around your use case instead of baking it into Yarn.

E.g. use a custom cache folder and remove it after install.

brandonsturgeon:
My use case here is to make local development easier when working with a project which uses a local folder as a package.

eg:
```javascript
// package.json
dependencies: {
    "my-dependency": "file:../my-dependency"
}
```

If I make a change in `../my-dependency`, I have to rebuild, remove the yarn.lock file, and delete `node_modules/my-dependency`, and then re-run `yarn install` or `yarn add file:../my-dependency`.

Sometimes I get an older version of `my-dependency` becuase it's being stored in the cache.

Maybe there's a better way to deal with this that I'm not aware of, but it would be great to be able to run `yarn install --no-cache` or `yarn add --no-cache file:../my-dependency`.

Just my two cents!

bestander:
@brandonsturgeon, your case is better handled by `knit` RFC https://github.com/yarnpkg/rfcs/pull/73 or `link:` specifier that is already available.

bestander:
Looks like there is some interested in this feature, spawning @BYK

nischayv:
+1 for this feature. Same reason as @brandonsturgeon

izonder:
+1 for the feature

One more case: I have a package.json
```
{
...
  "dependencies": {
    ... //only public npm packages
  },
  "devDependencies": {
    "pack": "git+ssh://git@gitlab.local/maintain/pack.git",
   ...
  }
  ...
}
```
yarn.lock contains the only one record with git+ssh:
```
"pack@git+ssh://git@gitlab.local/maintain/pack.git":
  version "1.0.0"
  resolved "git+ssh://git@gitlab.local/maintain/pack.git#8932fcdd220f4ae1d38a071ad8ffbcc552b45171"
```

Dockerfile contains:
```
...
RUN -x \
    && yarn install --production
...
```
But due to caching everything yarn requires `git` and `openssh` to be installed as well.

SmilingNavern:
+1 for this feature. I would prefer to disable cache completely, it can improve speed of our CI builds in docker and this cache is irrelevant because we don't need it at all.

herebebeasties:
Being able to do this via config (and therefore an env var) would be very beneficial for CI, not only for performance but also because of yarnpkg/yarn#683 (it's hard to mandate/educate that all the users of our CI server should run yarn with `--mutex network`, and that reduces concurrency anyway so is really sub-optimal). Being able to disable the cache entirely would effectively fix that issue.

csvan:
We also use Yarn in a Docker CI environment where cache is only overhead, since it is discarded as soon as the given build step is done. Being able to somehow disable cache would be greatly beneficial.

bestander:
I think there is enough interest in this feature.
Anyone wants to give it a try?

gaui:
This is such a basic crucial feature. This makes me want to switch to npm again.

dguo:
@bestander, do you have any high level suggestions for how someone might get started on implementing this?

bestander:
@dguo, it might be tricky, yarn's cache paths are deeply ingrained in fetching and linking phases.
At least yarn does not need to fetch .tar.gz files to read the dependency tree.

1. I would experiment with a specialized fetcher that saves to node_modules in the correct path before trying to add this feature to the existing fetcher.

2. I would consider some compromise.
Maybe still use cache folder but make it local inside, e.g. `node_modules/.cache`,  and then hard-link all packages from `node_modules/.cache` to appropriate `node_modules/module-a` folders.

evasyuk:
why it is not merged? it is does not have any collision with master

arcanis:
Because it's an rfc and it likely won't be implemented due to the internal design of the v1.

From the v2 onward there are no node_modules, the cache is the only artifact needed to run the project, so this RFC doesn't make sense there either.

jplwood:
@arcanis would you mind linking to some docs/discussion describing the v2 differences for those of us casual users popping in here interested in having the no-cache feature? Just curious to read up a bit more so i understand where yarn is going -- thanks!

arcanis:
The v2 leverages [PnP](https://yarnpkg.github.io/berry/features/pnp/) and zip loading to directly load the vendor files from within their archives (similar to phar archives in PHP, or asar in Electron). In this context the cache, albeit necessary, is the only artifact we generate. If you configure it to be stored within your project and don't version it, it will essentially be a "no cache Yarn".

adrian-skybaker:
https://yarnpkg.com/configuration/yarnrc#enableGlobalCache

oliviermartin:
I have a simple case why I would also need this 'no cache' option. Here is my react native project

```
my_new_react_component
├── ios
├── android
├── js
└── example   # My example application
                ├── ios
                ├── android
                └──  js
```

React Native does not support symlink in `node_modules`: https://github.com/facebook/metro/issues/1. So for installing my component in my example application (which is a subdirectory of my component), I use `yarn add` everytime I need to sync `my_new_react_component` in `example` app.
And because `example` is a sub-directory it copies even the example itself, its generated files/binary.

Note: I have not checked why but calling multiple times `yarn add ../` seems to consume more cache everytime - so the cache does not really work in this case.

An alternative would be a mechanism like `.npmignore` for Yarn for me to exclude `example` from my node module itself, but it does not seem to be supported: https://stackoverflow.com/questions/60088936/does-yarn-ignore-npmignore

PR #55, opened 2017-03-31
Yarn create

arcanis:
cc @cpojer @bestander - also cc @gaearon since create-react-app would probably be interested

### DIFF
--- a/text/0000-yarn-create.md
+++ b/text/0000-yarn-create.md
@@ -1,0 +1,67 @@
+- Start Date: 2017-03-31
+- RFC PR: (leave this empty)
+- Yarn Issue: (leave this empty)
+
+# Summary
+
+The idea would be to make it easier for users to setup projects from scratch.
+
+# Motivation
+
+During the past few years, we've seen an increase of the number of "boilerplate" projects, each one aiming to lower the complexity of creating new projects with Javascript. Create React App is a good example, but we can also mention Neutrino, released by the Mozilla teams, or Next.js, which recently reached its v2 milestone.
+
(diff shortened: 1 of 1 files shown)


bestander:
I think it is a good idea.
Would it work with most common bootstrappers (react, react native, ember, angular) out of the box?

arcanis:
I think so, maybe with two small restrictions:

- Semantically speaking, it would probably be for the best if they could support the `yarn create <package-name>` command without requiring extra parameter (for example, `ng new` doesn't seem to work, but `ng new foobar` does). I mean, it would work since we would forward the extra arguments to the create script anyway, but it would only fragment the syntax.

- They would have to support generating files into another directory than the current one (or they can do `(cd $npm_create_path && ng new)`, but it might not work on Windows ... I'm not sure about this one)

text/0000-yarn-create.md
```diff
@@ -0,0 +1,67 @@
...
+
+# Detailed design
+
+This RFC suggests to add a new command:
+
+```
+$> yarn create <package-name>
+```
+
+Running this command would have the samne effect as:
```
cpojer: "same"

text/0000-yarn-create.md
```diff
@@ -0,0 +1,67 @@
...
+
+```
+$> yarn add --dev <package-name>
```
cpojer: should be `yarn global add`, right?

arcanis: Maybe not, cf the third point https://github.com/arcanis/rfcs/blob/c6afb08a2a516c5dca4d03429c31bd8f91a9a03f/text/0000-yarn-create.md#alternatives

arcanis:
I've just published a tentative implementation on https://github.com/yarnpkg/yarn/pull/3244.

PR #56, opened 2017-04-18
Moves files around

arcanis:
This PR splits the `text` directory into `accepted` and `implemented`. I also updated the RFCs to include their PR & issue links, and potentially their implementation links.

cc @bestander


bestander:
Awesome work!
Thanks!

PR #57, opened 2017-04-23
Automatic Unit Testing During Upgrade Interactive

gburtini:
[Rendered](https://github.com/gburtini/rfcs/blob/75de75f2a15b9955fd10a617b28bf0a00566d27e/text/0000-unit-testing-interactive-upgrade.md)

The plan needs some development, but before I invest too much effort in it, I want to open the discussion to ensure there's not something obviously unreasonable about this. 

I am happy to put more effort in to a lot more specificity and perhaps even begin mocking up a version of the implementation if there's interest.

p.s., this idea was inspired by @cpojer and @bestander's talk about yarn and jest at F8. Thanks guys!

### DIFF
--- a/text/0000-unit-testing-interactive-upgrade.md
+++ b/text/0000-unit-testing-interactive-upgrade.md
@@ -1,0 +1,49 @@
+- Start Date: 2017-04-22
+- RFC PR:
+- Yarn Issue:
+
+# Summary
+
+Have `upgrade-interactive` run unit tests optionally to identify which upgrades
+are safe to run.
+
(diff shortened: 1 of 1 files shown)


bestander:
Thanks, @gburtini, especially that you were inspired by the talk.
This looks nice but I am afraid this will be hard to make generic and fast to be useful.

bestander:
This RFC has been open for quite, the idea is good but I think it goes beyond package manager responsibilities.
Probably a tool that wraps Yarn could do that

PR #58, opened 2017-04-29
CLI audio alerts

js-choi:
This RFC proposes that the Yarn CLI supports alert sounds for completion, warning, and error events. One could think of them as audio emoji. It could be done in a lightweight fashion with huge usability benefits.

There are some unfinished sections near the end about specific options for audio libraries, as well as some other details.

### DIFF
--- a/text/0000-cli-audio-alerts.md
+++ b/text/0000-cli-audio-alerts.md
@@ -87,3 +87,6 @@ These alternatives are not mutually exclusive with this RFC. They may deserve th
 
 ## Alternatives in CLI options
 **TBD**: Hierarchial/numeric alert levels
+
+# Unanswered questions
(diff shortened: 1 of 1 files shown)


bestander:
Thanks for sending the RFC.
Sounds (pun intended) like a fun feature and may be useful.

Does it have to be anything as complex as playing mp3?
Can Yarn produce ascii symbols that produce a ding sound when operation finishes?

bestander:
I think it is a cool idea but I don't think Yarn should have the complexity of music playback.
Cool features like that should exist in user land and Yarn should just allow people to plug in their types of Reporter class.

Another RFC that would allow people to override ConsoleReporter with custom behavior (e.g. playback) - that would be great.

PR #61, opened 2017-05-05
Adds workspace command

arcanis:


accepted/0000-workspaces-command.md
```diff
@@ -0,0 +1,35 @@
...
+$> (cd $(yarn project <project-name> pwd) && yarn <command-name> ...)
+```
+
+# Drawbacks
+
+- We will still have the issue of requiring the `--` separator to forward any command line option (`yarn project test install -- --production`). It's an important issue, something we really should tackle sooner than later.
+
+# Alternatives
+
+- Is `yarn project` a good name? I considered `yarn workspace`, but we're not operating on the workspace itself, just its projects. I feel like a `yarn workspace` command should only affect the root package.json (the one who contains the `workspaces` directive).
```
bestander: We planned to use `workspaces` keyword in package.json to identify all sub projects.
I would also try to stay away from term `project` because this term is overused in IT

thejameskyle:
See `yarn exec` and `yarn workspaces exec` / `yarn ws exec` in: https://github.com/yarnpkg/rfcs/pull/62

accepted/0000-workspaces-command.md
```diff
@@ -0,0 +1,35 @@
+- Start Date: 2017-05-04
+- RFC PR:
+
+# Summary
+
+This document specify a new command, that can be used to execute subcommands inside a workspace projects.
+
+# Motivation
+
+With the addition of workspaces, it will become handy to be able to execute commands inside other projects than the one in the current directory.
```
bestander: There are some specific scenarios that we need to address in this RFC.

For example, from my RFC #60

The structure of the source code is following

```
| jest/
| ---- package.json
| ---- packages/
| -------- babel-jest/
| ------------ packjage.json
| -------- babel-preset-jest/
| ------------ packjage.json
...
```

We want to address how we run commands (install/add/upgrade/remove) when CWD is workspace root.
Then what happens if CWD is jest/packages/babel-preset-jest?

How do I update a dependency in a workspace and in a root?

accepted/0000-workspaces-command.md
destroyerofbuilds: As if not overused already: `container`

bestander: I think we should limit the number of terms if possible.
I propose `workspaces root` and `workspaces`

accepted/0000-workspaces-command.md
arcanis: You can use `yarn add` to add a dependency in the current workspace, or `yarn workspace my-workspace add` to add a dependency inside a specific workspace from anywhere in the project tree.

arcanis:
Matching implementation: https://github.com/yarnpkg/yarn/pull/3365. Still a work-in-progress, but it seems to work relatively well. I'll clean it this afternoon, but still happy to hear feedbacks!

@thejameskyle I think the command described in this RFC is very similar to the one you described in your own RFC. Do you want to discuss here this specific part of the Workspace feature? I think we can reasonably also talk about your idea of `yarn workspace exec` :simple_smile:

accepted/0000-workspaces-command.md
arcanis: @thejameskyle suggested "project" for the workspaces root, and "workspace" for the ... workspaces.

Another option: "workbase" and "workset"? I'm not a native english speaker so maybe this doesn't make sense.

accepted/0000-workspaces-command.md
```diff
@@ -0,0 +1,31 @@
...
+
+With the addition of the Workspace feature, it will become handy to be able to execute commands inside other workspaces than the one in the current directory.
+
+# Detailed design
+
+This RFC suggests to add the two following commands:
+
+```
+$> yarn workspace <project-name> pwd
+$> yarn workspace <project-name> <command-name> ...
```
thejameskyle: One of the things I'd get a bit worries about here is overloading the same command space. There's three different things that fit into this space broken down into two groups:

- `yarn workspace <name> <command/run-script>` which matches `yarn <command/run-script>`
- `yarn workspace <name> <system-command>` (i.e. `yarn workspace <name> pwd`)

That's why I would like to see a new global command `exec` (which is similar to `lerna exec`)

| project-level | workspace-level |
| --- | --- |
| `yarn <command/run-script>` | `yarn workspace <name> <command/run-script>` |
| `yarn run <run-script>` | `yarn workspace run <run-script>` |
| `yarn exec <system-command>` | `yarn workspace exec <system-command>` |

arcanis: Hm, I see what you mean. The `pwd` command could be delegated to shell via `exec`. However, I think it would break the Windows portability. Since the command would run in cmd.exe, the pwd utility wouldn't be available, and users would have to use `cd` instead. In the end, the command would look like:

```
$> yarn workspace babel-core exec pwd || yarn workspace babel-core exec cd
```

Which looks awfully verbose (and not great perf-wise). Adding a builtin `pwd` command would solve this at virtually no cost, since we're need to compute the path anyway.

Regarding the `exec` command, I'm not sure how people will end up using it (they should probably register their commands as scripts if they intend to use them regularly, or to use advanced shell structures), but I have no objection implementing it, the maintenance cost shouldn't be very high anyway.

Just one concern: since Yarn already has a `run` command, aren't you worried that users might be confused regarding the command to use, depending on whether they want to run a script or a shell command?

thejameskyle: We've been using `lerna run` and `lerna exec` for a little while. I haven't heard of anyone confused by it. I think we can mitigate that with documentation if it does become an issue.

npm/Yarn `run` commands are already open to cross-platform incompatibilities since you could put `pwd` in `package.json#scripts` and have the same cross-platform problems. People (including me) still think its useful regardless.

thejameskyle: The other thing that we can do to make this useful cross-platform is to specify the `$PATH` the way `package.json#scripts` do.

From #62:

> There's three options for what the `$PATH` should be within workspaces.
> 
> - The project's `node_modules/.bin`
> - The workspace's `node_modules/.bin`
> - Both (w/ workspace before project)

arcanis: > We've been using lerna run and lerna exec for a little while. I haven't heard of anyone confused by it. I think we can mitigate that with documentation if it does become an issue.

Fine by me, then.

> The other thing that we can do to make this useful cross-platform is to specify the $PATH the way package.json#scripts do.

Ooooh, I like it, I often end up writing `"env": "env"` in my scripts just because I want to see the env my scripts will use. That could be an interesting use case for `yarn exec` (no workspace).

Ok, I'll update the document to add this.

accepted/0000-workspaces-command.md
```diff
@@ -0,0 +1,39 @@
...
+
+Recursion aside, it's essentially an alias for:
+
+```
+$> (cd $(yarn workspace <workspace-name> exec pwd) && yarn <command-name> ...)
+```
+
+# Drawbacks
+
+- We will still have the issue of requiring the `--` separator to forward any command line option (`yarn workspace test install -- --production`). It's an important issue, something we really should tackle sooner than later.
```
cpojer: I think we need to work through this. Yarn's own first-class flags should never require `--` to be passed. If you run `yarn workspace test jest -- --verbose` it makes sense, but not if it is part of a Yarn command.

cpojer: I wouldn't block this RFC on it though, I'm just saying I agree we need to make this work.

thejameskyle:
Do we want to limit it to valid set of sub-commands that Yarn will run inside workspaces?

arcanis:
Implementation-wise I don't think it's necessary, and I'd prefer to avoid maintaining a whitelist. Some commands will make little sense in a workspace context, sure, but who knows, maybe people will come up with some creative uses.

Ok! I'll merge this this afternoon if everyone's ok with this, and finish the PR in the train 👍

jardakotesovec:
Its not clear to me if is possible to install only subset of workspaces. `yarn install` in root folder install all workspaces. But in some cases (deployment of individual workspaces) we would like to install one or few of the workspaces, but still that it would install to the `node_modules` in top folder and would be merged with top package.json. 

Could be such use case considered?

arcanis:
I think we eventually will, but right now installing a worktree will install the dependencies for all the workspaces at once.

PR #62, opened 2017-05-05
[RFC] Yarn Workspaces

thejameskyle:
[[Rendered](https://github.com/thejameskyle/rfcs-1/blob/workspaces/accepted/0000-workspaces.md)]

### DIFF
--- a/accepted/0000-workspaces.md
+++ b/accepted/0000-workspaces.md
@@ -23,12 +23,6 @@ it would encourage more tools to abstract core functionality out.
 
 # Detailed design
 
-This is the bulk of the RFC. Explain the design in enough detail for somebody
-familiar with Yarn to understand, and for somebody familiar with the
-implementation to implement. This should get into specifics and corner-cases,
-and include examples of how the feature is used. Any new terminology should be
-defined here.
-
 ## Configuration
 
 The top-level project `package.json` may specify a `"workspaces"` field which
--- a/accepted/0000-workspaces.md
+++ b/accepted/0000-workspaces.md
@@ -1,0 +1,413 @@
+- Start Date: (fill me in with today's date, YYYY-MM-DD)
+- RFC PR: (leave this empty)
+- Yarn Issue: (leave this empty)
+
+# Summary
+
+Workspaces adds support for managing multiple packages within a single Yarn
+project. Linking between them on install to make cross-development simpler.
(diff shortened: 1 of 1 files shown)


bestander:
Hey @thejameskyle, that is a nice document you've had hidden from us :)

We've been considering trade offs and implementation details for months and decided to go with the phased approach. It's too large for one person to claim ownership over the full feature and we invite you to participate in the phased approach.

I have set up phases in this issue https://github.com/yarnpkg/yarn/issues/3294.
If you like you can claim a **piece** of it from RFC to full implementation.

thejameskyle:
I'm aware of that, I was in those meetings early on 😛 

There's nothing about this that cannot be broken down into phases, but the feature should really be considered all at once.

Also, I shared the private gist with you and @cpojer months ago

torifat:
> Right now we use a question field to manually type in a version. However, we should change that to a multi-choice selector (See Inquirer.js).

Recently I [moved](/yarnpkg/yarn/pull/3207) `Inquirer` inside our `ConsoleReporter`. I was thinking about refactoring the question/select. It will improve DX of several workflows.

accepted/0000-workspaces.md
```diff
@@ -0,0 +1,466 @@
...
+pieces of tools into their own packages because it would make development
+harder.
+
+If Yarn had a way of developing many packages as a single project which removed
+the additional maintenance cost of being able to test changes across packages,
+it would encourage more tools to abstract core functionality out.
+
+# Detailed design
+
+This is the bulk of the RFC. Explain the design in enough detail for somebody
```
JulianLeviston: Does this section need to be deleted, or filled in? Looks like boilerplate.

bestander:
Hey, @thejameskyle, thanks for sending this RFC, very inspiring.

Currently we think that we can address specifics faster in a phased approach that I track in https://github.com/yarnpkg/yarn/issues/3294.

Having focused discussions around particular isolated features allows us to have a faster turnaround and also experiment with code as the discussions happen.

Your spec has a good coverage around publishing experience and you have plenty related experience from lerna, do you want to champion that part of workspaces from RFC to implementation?

cpojer:
Closing this PR as it was superseded by the RFCs for Workspaces phase 1, 2 and 3.

PR #65, opened 2017-05-14
add alias system

voxsim:
After digging in many issues of the yarn project, I think we should do something like this!

Let me know if you like the idea ;)

### DIFF
--- a/accepted/0000-alias.md
+++ b/accepted/0000-alias.md
@@ -8,11 +8,8 @@ Add an alias system similar to `git alias` to the yarn project.
 
 # Motivation
 
-Everyone has different needs, It is right to don't have multiple commands that
-does the same thing, but at the same time I think it is right to give the possibility
-to the user to config their aliases. One example is that I prefer type `git dc`
-instead of `git diff --cached`, it is an habit, I know that and I will never ask the
-git project to support the `dc` alias.
+Everyone has different needs, It is right to don't have multiple commands that does the same thing, but at the same time I think it is right to give the possibility to the user to config their aliases. One example is that I prefer type `git dc`
+instead of `git diff --cached`, it is an habit, I know that and I will never ask the git project to support the `dc` alias.
 
 # Detailed design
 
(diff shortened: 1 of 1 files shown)


bestander:
Not a huge fan, I think modern shell interpreters can fill this niche without any change needed in Yarn.

gsklee:
This can be done in shell fairly easily: http://stackoverflow.com/questions/7131670/make-a-bash-alias-that-takes-a-parameter

voxsim:
Yep I know, but before closing this pr, I think we should decide some things:
- We now have a bunch of unsopported aliases, maybe we should delete that part (https://github.com/yarnpkg/yarn/blob/master/src/cli/unsupported-aliases.js) because it gives confusion in some way
- We sometime use ls, rm for some commands, maybe we should use always the long version and deprecate the short one. This is always for consistency.
- We should close all the issue related to alias something and tell everyone that we don't want to support this and I am ok with that, I can do without any problems :)

Let me know what you think ;)

gsklee:
Yeah, basically what we need is some valiant soul willing to do the triage works to clean things up a bit :p

bestander:
Ho ho ho, a new triage session is going to start!

bestander:
I've the next 2 weeks for issue triaging, this task is there for coordination https://github.com/yarnpkg/yarn/issues/3474.

The goal is to get ALL issues under control, prioritize and fix the ones critical for 1.0 release.

bestander:
Back to this PR, I vote to remove the aliases and educate people to create their own aliases for their workflows

gsklee:
Probably also wanna add a list of popular feature requests that have been discussed thoroughly and rejected into the issue template of either the main repo or this one.

bestander:
As for education, anyone, send a blog post entry here https://github.com/yarnpkg/website/tree/master/_posts how people can add aliases for their favorite operating system using Yarn

gsklee:
@bestander Finally got the time... once the blog post is accepted this RFC can be resolved I believe.

PR #66, opened 2017-05-18
Workspaces phase 3: linking

bestander:
Linking between workspaces

[Rendered](https://github.com/bestander/rfcs/blob/workspaces-phase-3/accepted/0000-workspaces-link-phase-3.md)


bestander:
Asking for comments from @arcanis @thejameskyle @cpojer @zertosh @Daniel15

accepted/0000-workspaces-link-phase-3.md
```diff
@@ -0,0 +1,164 @@
...
+    "type": "git",
+    "url": "https://github.com/facebook/jest.git"
+  },
+  "license": "BSD-3-Clause",
+  "main": "build/index.js",
+  "browser": "build-es5/index.js",
+  "dependencies": {
+    "chalk": "^1.1.3",
+    "diff": "^3.2.0",
+    "**jest-matcher-utils**": "^20.0.3",
```
sejoker: doesn't render as bold.

thejameskyle:
This is a bit separate, but one of the important things to keep in mind about the linking phase is what order you run the `postinstall/prepublish` scripts. It's possible for one workspace to depend on another another workspace within the same project and to use it as part of these scripts.

If you run the scripts in the wrong order on the workspaces, one might not be ready by the time another one goes to use it.

Lerna sorts them topologically based on which workspaces depend on one another. But that still has problems with circular dependencies.

I've been thinking about ways to mitigate this.

1. Give different priorities to dependency sorting so that "devDependencies" are treated more important for build scripts than "dependencies"
2. Add a way of resolving circular dependencies either as an additional field in workspace's `package.json` or in the project's `package.json`
3. Allow the Project's `package.json#workspaces` field specify dependencies in the order they should be built in:

```js
{
  "workspaces": [
    "packages/build-tool",
    "packages/*"
  ]
}
```

We could also combine some of these, they could all work together.

But the sorting process should definitely alert the user if it is unable to sort the workspaces definitively.

accepted/0000-workspaces-link-phase-3.md
bestander: Thanks, apparently I can't mix markup into code blocks

bestander:
Good points, @thejameskyle.
I'll add this to the RFC

accepted/0000-workspaces-link-phase-3.md
```diff
@@ -0,0 +1,164 @@
...
+A continuation of https://github.com/yarnpkg/rfcs/pull/60.
+Ability for workspaces to refer each other when testing packages in integration.
+
+## Motivation
+
+People tend to split larger projects into self contained packages that are published to npm independently. The workspaces feature is being developed for Yarn to address this workflow.
+
+In particular, testing packages that refer other packages from the same codebase can be difficult because Node.js and front end bundling tools would look up the referred packages in node_modules folder as it should be installed from npm registry.
+
+Yarn Workspaces need to enable referring other local packages the same way when local packages are in development mode (source of truth is the package source code) and in production mode (source of truth is the package installed from npm).
```
cpojer: nit: "need to be able to refer to other local packages"

accepted/0000-workspaces-link-phase-3.md
```diff
@@ -0,0 +1,164 @@
...
+## Detailed design
+
+The structure of the source code is following
+
+```
+| jest/
+| ---- package.json
+| ---- packages/
+| -------- jest-matcher-utils/
+| ------------ packjage.json
```
cpojer: package.json here and below.

accepted/0000-workspaces-link-phase-3.md
```diff
@@ -0,0 +1,164 @@
...
+  "workspaces": [
+    "packages/*"
+  ]
+}
+```
+jest-matcher-utils (workspace referred by another one)
+```
+{
+  "name": "jest-matcher-utils",
+  "description": "A set of utility functions for jest-matchers and related packages",
```
cpojer: Can you get rid of all the non-essential stuff from this package.json and replace it with `…`?

accepted/0000-workspaces-link-phase-3.md
```diff
@@ -0,0 +1,164 @@
...
+| ---- packages/
+| -------- jest-matcher-utils/
+| ------------ node_modules/ (empty, all dependencies hoisted to the root)
+| ------------ package.json
+| -------- jest-diff/
+| ------------ node_modules/
+| ---------------- **jest-matcher-utils**/  (symlink) -> ../jest-matcher-utils
+| ------------ package.json
+...
+```
```
cpojer: nice description of the solution.

accepted/0000-workspaces-link-phase-3.md
```diff
@@ -0,0 +1,164 @@
...
+
+### Dependencies and version matching
+
+Yarn would only link workspaces to each other if they match semver conditions.
```
cpojer: As @arcanis pointed out, we should absolutely warn in the case that it is not. In Jest, for example, this has caused issues in the past and we didn't know about it. Can you add it to the RFC?

accepted/0000-workspaces-link-phase-3.md
```diff
@@ -0,0 +1,164 @@
...
+| -------- diff/
+| -------- pretty-format/
+| ---- package.json
+| ---- packages/
+| -------- jest-matcher-utils/
+| ------------ node_modules/ (empty, all dependencies hoisted to the root)
+| ------------ package.json
+| -------- jest-diff/
+| ------------ node_modules/
+| ---------------- **jest-matcher-utils**/  (symlink) -> ../jest-matcher-utils
```
cpojer: Can you write down in words that the symlink is relative, just to make sure we have this captured properly?

accepted/0000-workspaces-link-phase-3.md
```diff
@@ -0,0 +1,164 @@
...
+
+
+### Problems with peer dependencies and hoisting
+
+There is a common [peer dependency problem](http://codetunnel.io/you-can-finally-npm-link-packages-that-contain-peer-dependencies/) when using **yarn link** on local packages that people can work around in Node 6+ by setting **--preserve-symlinks** runtime flag.
+In Workspaces this situation won't be a problem because node_modules are installed in Workspace root and Node.js `require()` statements will resolve third-party peer dependencies by going up the folder tree and reaching the Workspaces' root node_modules.
+
+As long as **jest-matcher-utils** does not make relative requires via its parent folder, flag **--preserve-symlinks** won't be necessary.
+
+
```
cpojer: rm empty duplicate line.

accepted/0000-workspaces-link-phase-3.md
```diff
@@ -0,0 +1,164 @@
...
+## Drawbacks
+
+This solution creates a symlink inside node_modules of a Workspace package and symlinks have multiple drawbacks:
+
+* Symlinks are not supported in all tools (e.g. watchman)
```
cpojer: We can potentially punt on this for a while.

accepted/0000-workspaces-link-phase-3.md
```diff
@@ -0,0 +1,164 @@
...
+* Symlinks are not supported well in all OS and environments (Windows pre 10 Creative updated, Docker on SMB storage(?))
```
cpojer: Could we have an alternative mode, especially for CI, where we run "yarn build" and then "yarn pack" on every package and cp the build into node_modules instead of using symlinks? CI's don't need the symlinks because you aren't changing the code there. Symlinks are only really important during development.

accepted/0000-workspaces-link-phase-3.md
```diff
@@ -0,0 +1,164 @@
...
+* A symlink to **jest-match-utils** does not emulate actual installation of the package, it just symlinks to the package source code - no prepublish and postinstall lifecycle scripts are executed and no files are filtered (as done during publishing)
```
cpojer: You are using the wrong module name ;)

accepted/0000-workspaces-link-phase-3.md
```diff
@@ -0,0 +1,164 @@
...
+        * Hardlinks have limited support in Windows pre 10
+        * When new files are created/removed in **jest-matcher-utils** the hardlinks need to be regenerated, that may require file watching to get good developer experience otherwise developer needs to run yarn install on every significant change
+        * This does not simulate actual installation of the package as no prepublish and postinstall lifecycle scripts are executed
+
+Yarn Workspaces could implement all of the above linking strategies and give developers a choice which one to choose for their project.
+Or the alternatives could be merged in a single solution for isolated e2e testing.
+
+## Unresolved questions
+
+* Is there still an issue with Node resolving real paths in symlinked folders (https://github.com/nodejs/node/issues/3402)? I think if all node_modules are installed in the Workspace root which is a parent to all workspaces and there are no relative (via ../../...) requires between workspaces everything should be working fine.
```
cpojer: I don't think RFCs should have language like "I think" :D Mind making this more objective?

accepted/0000-workspaces-link-phase-3.md
```diff
@@ -0,0 +1,164 @@
...
+* Should the symlinks be created in Workspace root node_modules instead of referring Workspaces?
+* Any special treatment for scoped packages?
```
cpojer: Can you answer this? It seems easy to solve, they just need to be supported normally, no?

accepted/0000-workspaces-link-phase-3.md
```diff
@@ -0,0 +1,164 @@
...
+* Does it need to work for other type of packages: git, file, etc?
+* As described in Workspace phase 1 RFC (https://github.com/yarnpkg/rfcs/pull/60) there is only one lockfile per workspace. Should we have workspaces that are referred from other workspaces referred in the root lockfile?
```
cpojer: This is a good question. Do we only want to enable one gigantic workspace or should we enable slices of projects that compose different workspaces (this may not be possible)? For example, if you take Jest and react-native-packager, they have some overlapping modules: would you have two workspaces, one for each project where some modules are shared or would you have one workspace for both projects?

The reason I'm asking is because at FB, if we end up linking 100 packages together, that may actually be slower in the end and managing them outside a workspace may be better. Not something that blocks the current RFC and feature, but something to keep in mind.

cpojer:
Also, can you elaborate on the decision to deprecate `yarn knit` or do it in a separate PR? I don't think it fits into this PR.

accepted/0000-workspaces-link-phase-3.md
arcanis: Running yarn build implies that the module has its dependencies ready, which might not be the case when working with cyclic dependencies.

accepted/0000-workspaces-link-phase-3.md
thejameskyle: I don't think it should always warn. This breaks Babel's use case where it is self-hoisting and wants to use an older version intentionally.

accepted/0000-workspaces-link-phase-3.md
thejameskyle: I don't think there's any need to treat them specially. Maybe I'm missing something though

accepted/0000-workspaces-link-phase-3.md
bestander: I suppose it could report which packages are linked after install.
I agree with @thejameskyle that it can be annoying if it is intended

bestander: But I think it is early to spec this detail, we'll need to see what level of reporting works here.
Definitely it should be in verbose logs

accepted/0000-workspaces-link-phase-3.md
bestander: I've seen once a cyclic dependency with post install scripts.
Not a common but possible case.
What if we just don't support cyclic workspaces?
It seems like a poor design and causes a lot of trouble.

accepted/0000-workspaces-link-phase-3.md
bestander: I think it is correct.
jest-diff -> jest-match-utils

accepted/0000-workspaces-link-phase-3.md
bestander: Ok, just wanted to confirm it in first iteration

accepted/0000-workspaces-link-phase-3.md
cpojer: It's `jest-matcher-utils`.

accepted/0000-workspaces-link-phase-3.md
bestander: This is a different question, I'll rephrase it.
I think the one that you raised is out of scope for this RFC and we'll need to get back to it later.

accepted/0000-workspaces-link-phase-3.md
bestander: 300px-paris_tuileries_garden_facepalm_statue

bestander:
> Also, can you elaborate on the decision to deprecate yarn knit or do it in a separate PR? I don't think it fits into this PR.

I am rolling this back.
I thought that `yarn knit` would be deprecated by workspaces and linking but probably we would want to have both anyway and some code could be shared.

bestander:
> One of the larger issues I'm worried about is that we aren't discussing about build steps here. If we were to create a workspace that contains projects from two separate parent projects that have a different build pipeline, for example one of them uses TypeScript and the other one uses babel, then we don't know how to build the project. This doesn't need to be addressed now, but I believe this needs to be on our mind long term. Imagine Jest's watch mode, instead of running tests, watching for file changes and building different projects whenever something happens ;P

There are 2 aspects that I missed:

- build order of install scripts (addressed by @thejameskyle https://gist.github.com/thejameskyle/abbc146a8cb5c26194c8acc4d14e7c30)

- Sharing workspaces between multiple roots.
I would keep it out of scope for now.

PR #67, opened 2017-05-19
Add yarn scripts RFC.

cpojer:


bestander:
I think it is quite useful.
I think this can become a powerful feature in Yarn over time but I wonder if at this stage we should treat it as experimental, we might want to add breaking changes.

ljharb:
Why yaml over something more secure like toml?

rally25rs:
Typically if I have complicated scripts, I just include a `/bin` or `/scripts` directory and put bash or bat files in there.

i.e. in package.json

```
{
  "scripts": {
    "clean": "bin/clean.sh"
  }
}
```

Is this scripts redesign getting me something new that isn't easier to manage in a separate set of shell scripts?

Even in your example about Travis CI using yaml; in our builds we've just refactored it all into shell scripts, because escaping quotes can be painful in yaml (imagine nesting sed/awk and regexes in yaml)

For example my typical travis.yml has just become:
```
before_script:
- . ./travis_scripts/update_node_version.sh
script:
- . ./travis_scripts/build_ui.sh
```

Not trying to discredit the idea, just looking for the "killer feature" / differentiator.

-----

Perhaps there is opportunity here: complication often comes from managing shell scripts for win vs linux vs osx. Could the yarn-scripts.yml have an OS-specific layer of abstraction? Like `yarn run foo` could run a bash script on linux and a .bat file on Windows.

gsklee:
I agree with @rally25rs, maybe I'm missing the real point but I failed to see it. The most widely adopted strategy in managing npm scripts right now is to use separate shell scripts as @rally25rs has mentioned, and it doesn't have to discriminate between `install`-related scripts and others:

```json
{
	"scripts": {
		"preinstall": "./commands/preinstall",
		"install": "./commands/install",
		"test": "./commands/test"
	}
}
```

I haven't heard of `nps`, just checked it out and I think it's borderline overengineering. Also, the use of YAML file format is unjustified. Unlike JSON which is native to the JS environment and Shell scripts which are native to CLI, YAML is a foreign format; it can very well be TOML or other INI-like formats. This can lead to a few problems:

* Bloat: Yarn will now need to carry a foreign format parser around just to be able to read this "nice-to-have" optional file.

* Bad precedent: Representatives from [all other data serialization formats](https://en.wikipedia.org/wiki/Comparison_of_data_serialization_formats) can now chime in and demand their better-than-JSON format be added since YAML/TOML/whatever has been added already. More bloat requests to deal with.

* Compatibility with npm users: For npm users to be able to use npm scripts, the `scripts` field needs to remain intact as-is. So package authors now need to ship the information in two places, inside package.json's `script` field and another separated YAML/TOML/whatever file. More bloat.

So unless there is some hidden benefit that is yet to become apparent, I am not throwing my support behind this proposal.

kittens:
This is just reinventing `make`. Change the ` - ` to `\t` and it'd be a valid Makefile.

Daniel15:
> Could the yarn-scripts.yml have an OS-specific layer of abstraction? Like yarn run foo could run a bash script on linux and a .bat file on Windows.

I wonder how much of an issue that really is. In some projects (like Yarn itself) I just use a shell script even on Windows. A lot of Windows users have some port of Bash installed (Git comes bundled with one) which can be used to run shell scripts. Our Windows builds run `scripts/build-dist.sh` and things mostly work. On some of my projects, I go the other way and use PowerShell for the scripts, since it's cross-platform and pretty powerful.

Slightly off-topic but you'd want to use PowerShell on Windows rather than batch files. Batch files are only used for super simple scripts, anything more than a few lines would likely greatly benefit from PowerShell 😛

arcanis:
I'd prefer a makefile-like format: http://peg.arcanis.fr/23THQ1/

Makefile isn't perfect, but it express quite well the relations between the rules.

cpojer:
I'll close this for now and come back with a different idea. @byk convinced me that we shouldn't extend the task running capabilities in Yarn.

PR #68, opened 2017-05-21
Selective versions resolutions

victornoel:
Hi,

Here is an RFC for solving the situation discussed in yarnpkg/yarn#2763

@bestander

### DIFF
--- a/accepted/0000-selective-versions-resolutions.md
+++ b/accepted/0000-selective-versions-resolutions.md
@@ -208,19 +208,21 @@ latter is the result of yarn resolution being influenced by the *resolution*).
 dependency `a`.
 - `**/a/b` denotes the directly nested dependency `b`
 of all the dependencies and nested dependencies `a` of the project.
-- `a` denotes the project's dependency `a` (see below for a discussion on
-the matter of specifying a *resolution* for one of the project dependency).
 - `a/**/b` denotes all the nested dependencies `b` of the project's
 dependency `a`.
 - `**/a` denotes all the nested dependencies `a` of the project.
 - `a` is an alias for `**/a` (for retro-compatibility, see below, and because
 if it wasn't such an alias, it wouldn't mean anything as it would represent
 one of the non-nested project dependencies, which can't be overridden as
-explained below)
-- `**/a-*` denotes all the nested dependencies of the project whose
-name starts with `a-`.
+explained below).
 - `**` denotes all the nested dependencies of the project (a bad idea mostly,
 as well as all other designations ending with `**`).
+
+Note on single star: `*` is not authorized in a package resolution because it
+would introduce too much non-determinism. For example, there is the risk of a
+referring to `package-*` at one point to match `package-a` and `package-b`,
+and later on, this would match a new nested dependency `package-c` that wasn't
+intended to be matched.
 
 ### Version specification
 
@@ -258,8 +260,15 @@ as expected: the nested dependencies will have the fixed version specified.
 
 ## Relation to the `--flat` option
 
-The `--flat` option becomes thus a way to populate the resolutions field for
-the whole project, as it already does (using a package designation in the
+Before this RFC, `--flat` is both about populating resolutions field AND
+taking resolutions field into account when executing the `install` command
+(including installation as part of the `add` command).
+
+This RFC is about taking the `resolutions` field into account when executing
+the `install` command  (including installation as part of the `add` command).
+
+So with this RFC, `--flat` is now only about populating resolutions field.
+I does it in the same way as before (using a package designation in the
 form of `package-name`).
 
 The only breaking change is that the `resolutions` field is always considered
@@ -275,8 +284,8 @@ installation itself, not the adding), it will continue to behave as before
 by asking for *resolutions* of all the nested dependencies of the project even
 with `add`.
 
-See open question below about the need for removing `--flat` from `install`
-and for introducing a `flatten` command instead.
+In the future, `--flat` will need to be rethought but for now we will keep
+its behaviour.
 
 ## `yarn.lock`
 
@@ -342,6 +351,11 @@ be fixed at one point or another and the resolution should be removed.
 In that sense, incompatible resolutions should always be warned about.
 Furthermore, an incompatible resolution is a potential for unwanted behaviour
 and should thus never be ignored by the user.
+
+## Locality of the *resolutions*
+
+The `resolutions` field only apply to the local project and not to the projects
+that depends on it. It is the same as with lock files in a way.
 
 # How We Teach This
 
@@ -459,21 +473,12 @@ The two alternatives discussed in the section just above, "Mapping version
 specifications" and "Mapping version specifications as well as packages name",
 can be adapted to the current proposition to support these uses cases as well.
 
-# Unresolved questions
+## `flatten`
 
-## Package designation and single `*`
+Some notes on `--flat` and its future with respect to this RFC.
 
-I am not totally convinced that using a single `*` in a package designation
-is a good idea. Mostly because it introduces much for uncertainty to what it is
-going to be applied. In particular, I'm thinking about defining a *resolution*
-like that and latter adding a dependency which has a dependency that matches
-but is not intended to. I feel like it adds a lot of mental weight for the
-user to manipulate single `*`.
-
-## `--flat` option
-
-I wonder if the `--flat` option of `install` shouldn't be transformed to
-a `flatten` command that would:
+The `--flat` option of `install` could be transformed to a `flatten` command
+that would:
 1. Fill in the *resolutions* for all nested dependencies.
 2. Set the `flat` field in the `package.json`.
 
@@ -482,14 +487,7 @@ It makes no real sense to have a flattening mode for `install`:
 2. `install` should be only about building the `node_modules` directory, not
 modifying the the `package.json` IMHO.
 
-If we do that, then the `--flat` option of `add` (and the `flat` option in the
-`package.json`) would apply not to the installation but to the adding,
-upgrading, etc (everything that modify the `package.json`'s dependencies):
-it will ensure that the project stays flattened via the populating of the
-`resolutions` field.
-
-## resolutions in dependencies
-
-If a dependency contains resolutions, should it be taken into account?
-In Maven, it is not the case for example.
-In other words: should we restrict resolutions to `private` project?
+Then the `flat` option in the `package.json` (and the `--flat` option of `add`)
+would apply not to the installation but to the adding, upgrading, etc
+(everything that modify the `package.json`'s dependencies). It will ensure
+that the project stays flattened via the populating of the `resolutions` field.
--- a/accepted/0000-selective-versions-resolutions.md
+++ b/accepted/0000-selective-versions-resolutions.md
@@ -317,7 +317,7 @@ seen it happen many times with maven, which is quite complex but complete in
 its dependency management. Users would get confused and it can take time to
 understand the implications of manipulation the `resolutions` field .
 
-# Alternatives and/or extensions
+# Alternatives
 
 ## Global nested dependencies resolution
 
--- a/accepted/0000-selective-versions-resolutions.md
+++ b/accepted/0000-selective-versions-resolutions.md
@@ -85,9 +85,12 @@ file to be considered all the time and on a per-package basis.
 When a nested dependency is being resolved by yarn, if the `resolutions` field
 contains a version for this package, then this version is used instead.
 
-All the examples are given with exact dependencies, but note that putting a
+Most of the examples are given with exact dependencies, but note that putting a
 non-exact specification in the `resolutions` field should be accepted and
-resolved by yarn like it usually does.
+resolved by yarn like it usually does. This subject is discussed below also.
+
+Any potentially counter-intuitive situation will result in a warning being
+issued. This subject is discussed at the end of this section.
 
 ## Example
 
@@ -136,13 +139,35 @@ provoke the modification of the `yarn.lock` (see
 
 This feature would remove the need for this behaviour of yarn.
 
+## Non-exact version specifications
+
+If there is a non-exact specifications in the `resolutions` field, the rule is
+the same: the `resolutions` field takes precedence over the specification in a
+nested dependency.
+
+In case the `resolutions` field is broader than the nested dependency
+specification, then a warning can be issued. This happens if the the exact
+version resolved by yarn based on the `resolutions` specification is
+incompatible with the nested dependency specification.
+
+For example, if `@angular/cli` depends on `typescript@>=2.0.0 <2.3.0` and the
+`resolutions` field contains `typescript@>=2.0.0 <2.4.0`, then if the latest
+available version for typescript is `2.2.2`, no warning is issued, and if the
+latest available version for typescript is `2.3.2` then a warning is issued.
+
+The rational behind that is that since the `yarn.lock` file is only modified
+by the user (via yarn commands), then a warning will always be issued before
+such a situation happens and is written to the `yarn.lock` file.
+
 ## Warnings in logs
 
-yarn would need to warn about the following situations:
+yarn should warn about the following situations:
 - Unused resolutions
 - Incompatible resolutions: see the above section about `yarn.lock`.
 Incompatible resolutions should be accepted but warned about since it could
 lead to unwanted behaviour.
+- Broadening specifications: see above about non-exact specifications. This
+actually falls under the umbrella of incompatible resolutions.
 - ? (see open questions below)
 
 # How We Teach This
--- a/accepted/0000-selective-versions-resolutions.md
+++ b/accepted/0000-selective-versions-resolutions.md
@@ -50,8 +50,6 @@ It makes sense for typescript as the user intent is clearly to use typescript
 CLI (responsible of compiling `.ts` files) will simply use the 2.2.2 version
 from its `node_modules`.
 
-## Variations of the original problem
-
 Similarly, even using such a content for `package.json`:
 ```json
  "devDependencies": {
@@ -61,6 +59,8 @@ Similarly, even using such a content for `package.json`:
 
 The need could arise for forcing the use of `typescript@2.3.2` (or
 `typescript@2.1.0` for that matter).
+
+## Why?
 
 In these example, the need does not seem very important (the user could maybe 
 use `typescript@2.2.2` or ask the `@angular/cli` dev team to relax its
@@ -80,46 +80,158 @@ See alternatives solutions below also.
 # Detailed design
 
 The proposed solution is to make the `resolutions` field of the `package.json`
-file to be considered all the time and on a per-package basis.
+file to be considered all the time and on a per-package basis (instead of
+only when the `--flat` parameter is used).
 
 When a nested dependency is being resolved by yarn, if the `resolutions` field
-contains a version for this package, then this version is used instead.
+contains a specification for this package, then it will be used instead.
 
-Most of the examples are given with exact dependencies, but note that putting a
+Special attention to the specific nature of package management in the npm
+ecosystem is given in this RFC: indeed, it is not unusual to have the same
+package being present as a nested dependency of multiple packages with
+different versions. It is thus possible within the `resolutions` field to
+express versions either for the whole dependency tree or only for a subset of
+it, using a syntax relying on glob patterns.
+
+Most of the examples are given with exact dependencies, but note that using a
 non-exact specification in the `resolutions` field should be accepted and
 resolved by yarn like it usually does. This subject is discussed below also.
 
 Any potentially counter-intuitive situation will result in a warning being
 issued. This subject is discussed at the end of this section.
 
-## Example
+## Examples
 
-For example with:
+We the following packages and their dependencies:
+```
+package-a@1.0.0 and 2.0.0
+ |_ package-d1@1.0.0
+     |_ package-d2@1.0.0
+
+package-b@1.0.0
+ |_ package-d1@2.0.0
+
+package-c@1.0.0
+ |_ package-a@2.0.0
+     |_ ...
+```
+
+With:
 ```json
  "devDependencies": {
-    "@angular/cli": "1.0.3"
+    "package-a": "1.0.0",
+    "package-b": "1.0.0"
   },
   "resolutions": {
-    "typescript": "2.3.2"
+    "**/package-d1": "3.0.0"
   }
 ```
 
-yarn will use `typescript@2.3.2` for every nested dependency to `typescript`
+yarn will use `package-d1@3.0.0` for every nested dependency to `package-d1`
 and will behave as expected with respect to the `node_modules` folder by not
-duplicating typescript installation.
+duplicating the `package-d1` installation.
 
-## Relation to non-nest dependencies
+With:
+```json
+ "devDependencies": {
+    "package-a": "1.0.0",
+    "package-b": "1.0.0"
+  },
+  "resolutions": {
+    "package-a/package-d1": "3.0.0"
+  }
+```
+
+yarn will use `package-d1@3.0.0` oly for `package-a` and `package-b` will
+still have `package-d1@2.0.0` in its own `node_modules`.
+
+With:
+```json
+ "devDependencies": {
+    "package-a": "1.0.0",
+    "package-c": "1.0.0"
+  },
+  "resolutions": {
+    "**/package-a": "3.0.0"
+  }
+```
+
+`package-a` will still be resolved to `1.0.0`, but `package-c` will have
+`package-a@3.0.0` in its own `node_modules`.
+
+With:
+```json
+ "devDependencies": {
+    "package-a": "1.0.0",
+    "package-c": "1.0.0"
+  },
+  "resolutions": {
+    "package-a": "3.0.0"
+  }
+```
+
+yarn will do nothing (see below why).
+
+With:
+```json
+ "devDependencies": {
+    "package-a": "1.0.0",
+    "package-c": "1.0.0"
+  },
+  "resolutions": {
+    "**/package-a/package-d1": "3.0.0"
+  }
+```
+
+yarn will use `package-d1@3.0.0` both for `package-a` and the nested
+dependency `package-a` of `package-c`.
+
+## Resolutions
+
+Each sub-field of the `resolutions` field is called a *resolution*.
+It is a JSON field expressed by two strings: the package designation on the
+left and a version specification on the right.
+
+### Package designation
+
+A *resolution* contains on the left-hand side a glob pattern applied to
+the dependency tree (and not to the `node_modules` directory tree, since the
+latter is the result of yarn resolution being influenced by the *resolution*).
+
+- `a/b` denotes the directly nested dependency `b` of the project's
+dependency `a`.
+- `**/a/b` denotes the directly nested dependency `b`
+of all the dependencies and nested dependencies `a` of the project.
+- `a` denotes the project's dependency `a` (see below for a discussion on
+the matter of specifying a *resolution* for one of the project dependency).
+- `a/**/b` denotes all the nested dependencies `b` of the project's
+dependency `a`.
+- `**/a` denotes all the nested dependencies `a` of the project.
+- `**/a-*` denotes all the nested dependencies of the project whose
+name starts with `a-`.
+- `**` denotes all the nested dependencies of the project (a bad idea mostly,
+as well as all other designations ending with `**`).
+
+### Version specification
+
+A *resolution* contains on the right-hand side a version specification
+interpreted via the `semver` package as usually done in yarn.
+
+## Relation to non-nested dependencies
 
 The `devDependencies` and `dependencies` fields always take precedence over the
 `resolutions` field: if the user defines explicitly a dependency there,
 it means that he wants that version, even if it's specified with a non-exact
 specification. So the `resolutions` field only applies to nested-dependencies.
+Nevertheless, in case of incompatibility between the specification of a
+non-nested dependency version and a *resolution*, a warning is issued.
 
 ## Relation to the `--flat` option
 
 The `--flat` option becomes thus a way to populate the resolutions field for
-the whole project, as it already does.
-But the `resolutions` field is always considered by yarn, even if `--flat` is
+the whole project, as it already does (using a package designation in the
+form of `**/package-name`).
+And the `resolutions` field is always considered by yarn, even when `--flat` is
 not specified.
 
 Incidently, this resolves this strange situation when two developers would be
@@ -130,8 +242,8 @@ and they would get different `node_modules` contents because of that.
 
 This design implies that it is possible to have for a given version
 specification (e.g., `>=2.0.0 <2.3.0`) a resolved version that is incompatible
-with it (e.g., `2.3.2`).
-It is acceptable as long as it is explicitly asked by the user.
+with it (e.g., `2.3.2`). It is acceptable as long as it is explicitly
+asked by the user via a *resolution*.
 
 It is currently the case that such situation would make yarn unhappy and
 provoke the modification of the `yarn.lock` (see 
@@ -187,7 +299,13 @@ sole control of yarn itself anymore, but also under the control of the user
 now.
 
 This is an advanced use of yarn, so new users don't really have to know about
-it in the beginning.
+it in the beginning. Still, it is meant to be used on a potential regular
+basis, in particular when some packages a project depends on have problems
+in their how dependencies.
+
+Thus it would make sense to have a bit of the documentation talking about
+this use case and underlying the fact that *resolutions* are mostly here
+on a temporary basis.
 
 # Drawbacks
 
@@ -197,30 +315,11 @@ It makes yarn behaviour a bit more complex, even though more useful. So it
 can be difficult for users to wrap their head around it. The RFC submitter has
 seen it happen many times with maven, which is quite complex but complete in
 its dependency management. Users would get confused and it can take time to
-understand the implications of manipulation the `resolutions` field (even
-though, the chosen solution, compared to the alternatives below, is much
-simpler).
+understand the implications of manipulation the `resolutions` field .
 
-## Package management paradigm
+# Alternatives and/or extensions
 
-Yarn and npm users are highly used to the idea that a dependency can be
-present many times in the `node_module`, depending on which package needs it.
-This has advantages and disadvantages, but it is one of the specificity of the
-npm ecosystem package management.
-
-In this light, taking such as design decision puts yarn a bit farther to such
-way of doing thing, and it could be considered a bad direction to go toward.
-
-Some of the alternatives below actually take this into consideration, but are
-a bit more complex in terms of expressiveness, so were not chosen by the RFC
-submitter (see open questions below too).
-
-# Alternatives
-
-There is at least one alternative to the proposed solution, more complex but
-more expressive.
-
-## Nested dependencies resolution per dependency
+## Global nested dependencies resolution
 
 Starting from an example, this solution would take the following form in the
 `package.json` file:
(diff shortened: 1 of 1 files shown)


accepted/0000-selective-versions-resolutions.md
```diff
@@ -0,0 +1,278 @@
...
+- RFC PR: (leave this empty)
+- Yarn Issue: (leave this empty)
+
+# Summary
+
+Allow to select a nested dependency version via the `resolutions` field of
+the `package.json` file.
+
+ça veut dire qu'on doit pouvoir mettre des specs invalides (wrt vesion) dans
+lockfile.
```
xtuc: Why is this written in french? Should we translate this part?

victornoel: haha no, sorry, I took some notes while writing and forgot to remove them :P I'm going to amend the commit!

victornoel: It's done, thanks :)

accepted/0000-selective-versions-resolutions.md
```diff
@@ -0,0 +1,275 @@
...
+
+# Detailed design
+
+The proposed solution is to make the `resolutions` field of the `package.json`
+file to be considered all the time and on a per-package basis.
+
+When a nested dependency is being resolved by yarn, if the `resolutions` field
+contains a version for this package, then this version is used instead.
+
+All the examples are given with exact dependencies, but note that putting a
```
sejoker: it make sense to provide separate examples for the resolutions with non-exact specification.

accepted/0000-selective-versions-resolutions.md
```diff
@@ -0,0 +1,275 @@
...
+specification. So the `resolutions` field only applies to nested-dependencies.
+
+## Relation to the `--flat` option
+
+The `--flat` option becomes thus a way to populate the resolutions field for
+the whole project, as it already does.
+But the `resolutions` field is always considered by yarn, even if `--flat` is
+not specified.
+
+Inceidently, this resolves this strange situation when two developers would be
```
sejoker: Incidentally

accepted/0000-selective-versions-resolutions.md
victornoel: Ok, I will work on that soon then :)

My idea is that the way yarn resolves a given dependency is the same as before, so exact or non-exact specifications should be handled the same.
For example, the role of `yarn.lock` doesn't change, it is completely inferred from the `package.json` file, including the `resolutions` field.

But maybe there is some corner cases to be discovered. Im' thinking for example about a situation where a nested dependency has a specification more specific than what is specified in the `resolutions` field. Intuitively I would say that the `resolutions` field takes precedence and a warning is printed. I will put that in a more formal way in the document :)

sejoker:
@victornoel in https://github.com/yarnpkg/yarn/issues/2763 you mentioned what npm allows to override dependencies. Could you reference that in RFC? I can't find any documentation on npm website, maybe I missed something. The point is, referencing npm's similar mechanism will increase chances of the RFC to land.

accepted/0000-selective-versions-resolutions.md
victornoel: it's done :)

victornoel:
@sejoker actually I'm not so sure it is possible with npm, I will look a bit more, but what I originally linked (https://nodejs.org/en/blog/npm/managing-node-js-dependencies-with-shrinkwrap/) is actually closer to the `--flat`  options than to this RFC.

I still remember reading something about that but there is a lot of chance I was wrong.

accepted/0000-selective-versions-resolutions.md
```diff
@@ -0,0 +1,303 @@
...
+As explained in the alternative solutions section, it would be much more
+expressive and coherent with the npm ecosystem package management paradigm
+to use nested dependency resolutions per project dependency.
+Would the loss of simplicity acceptable maybe?
+
+## Warnings in logs
+
+Should yarn warn the user about an incoherence between an explicit dependency
+and a resolution. For example if the user specify a dependency to
+`typescript@2.3.2` and the resolutions field contains `typescript@2.3.0`.
```
bestander: Considering that Yarn warns when peer dependencies don't match semver I think overridden resolutions should warn as well.
I think this is a temporary measure until the community catches up and fixes issues

accepted/0000-selective-versions-resolutions.md
```diff
@@ -0,0 +1,303 @@
...
+For sure if the above alternative solution is chosen, this wouldn't make sense.
+
+Should we warn if a resolutions is incompatible, but still upper-bounded?
+For example, forcing version `a@2.3` while a dependency needs version `a@2.2`
```
bestander: I think if semver is violated there should be a warning.

accepted/0000-selective-versions-resolutions.md
```diff
@@ -0,0 +1,303 @@
...
+
+Starting from an example, this solution would take the following form in the
+`package.json` file:
+```json
+  "devDependencies": {
+    "@angular/cli": "1.0.3",
+    "typescript": "2.3.2"
+  },
+  "resolutions": {
+    "@angular/cli": {
```
bestander: I would prefer nesting to be an option.
JS project, unlike Java, have hundreds of dependencies, it is not rare to have several versions of the same library at different levels and this is fine.
Sometimes I'd need a fine control to override resolution only for a specific dependency path

bestander:
You need to consider `check --verify-tree` and `check` commands.
First one parses package.json files and verifies that they don't conflict with each other.

bestander:
Thanks for raising this, @victornoel!

arcanis:
Three thoughts:

- Maybe we could put this inside the lockfile instead of the package.json? I assume this `resolutions` key will have no effect unless it's inside the project root, right (just like we ignore nested `yarn.lock`)?

- Maybe we could make the `resolutions` key a glob pattern, minus the `node_modules`:

    ```json
    { "resolutions": { "**/@angular/cli/typescript": "ssh+git://..." } }
    ```

    It would be more verbose, but this feature should be a very advanced feature anyway, and should never be used under normal circonstances, so I think it could work.

- Maybe the `resolutions` could become a "preference" array, so that :

    ```json
    { "resolutions": { typescript: [ "ssh+git://..." ] } }
    ```

    Then, when resolving, we would first try the dependencies listed in this array, in the order defined. If they don't match the version specified, then continue just like we currently do. The problem would be to draft a good wording for the exotic resolves (ie dependencies that don't come from the npm registry).

bestander:
> Maybe we could put this inside the lockfile instead of the package.json?

I think this list will be manually managed and yarn.lock file is automatically generated.
Also IMHO resolutions is very similar dependencies from package.json, they loosely influence resolution step in Yarn install command, yarn.lock definitely describes resolved dependencies after resolution happened.

accepted/0000-selective-versions-resolutions.md
victornoel: yes, that's exactly why I am not so sure the original proposition is good enough and I added that question at the end :)

Should we merge both of the proposition?

Accepting both resolutions like this:
```
"typescript": "2.3.2"
```
and
```
"@angular/cli": {
    "typescript": "2.3.2"
}
```

victornoel:
@arcanis originally it is not intended to be a very advanced feature. It is meant to be used regularly by people. At least in the Maven world, this kind of feature is very useful and thus very used.

Your question raises a point I didn't think of before: is it the case that you can have `node_modules` directory inside dependencies that are themselves in the `node_modules` of one of the project dependency?
Something like `myproject/node_modules/dep1/node_modules/dep2/node_modules/dep3`?

bestander:
@victornoel, feel free to start work on the implementation, we can agree on the details how specifiers are implemented in the meantime.

bestander:
> Something like myproject/node_modules/dep1/node_modules/dep2/node_modules/dep3

That is a legit use case.

victornoel:
> feel free to start work on the implementation, we can agree on the details how specifiers are implemented in the meantime.

@bestander I'm not sure I have the skills to implement this, I wouldn't even know where to start…
I can commit to take some time to see how it could be implemented and see if I'm up to the task, but I would need you to give me some guidance on where to start for this to happen.

>> Something like myproject/node_modules/dep1/node_modules/dep2/node_modules/dep3

> That is a legit use case.

It is? Damn, it means that to be complete, we would need to be able to specify resolutions like this then:
```json
"resolutions": {
    "dep1": {
        "dep2": {
            "dep3": "1.0.0"
        }
    }
}
```

Or use some kind of glob pattern as proposed by @arcanis

bestander:
>  I'm not sure I have the skills to implement this, I wouldn't even know where to start…

Most of resolution stuff is happening here https://github.com/yarnpkg/yarn/blob/master/src/package-resolver.js.
You can also search where `semver` module is used because your change is going to be about violating semver.

> Or use some kind of glob pattern as proposed by @arcanis

For deep trees, which is common, a glob pattern makes sense.
A maybe easier alternative is to override a specific semver with another one, e.g. `typescript@>=2.5 < 3` => `typescript@4.0.2`

victornoel:
> A maybe easier alternative is to override a specific semver with another one, e.g. typescript@>=2.5 < 3 => typescript@4.0.2

I'm not very convinced by this approach, I explained why in the RFC (I will update it with all the feedbacks and discussions soon by the way).

bestander:
It just seemed easier to implement for me but I don't insist

victornoel:
> It just seemed easier to implement for me but I don't insist

I agree it will be easier to implement, I'm more worried about usability and expressiveness :)

accepted/0000-selective-versions-resolutions.md
victornoel: agree

accepted/0000-selective-versions-resolutions.md
victornoel: agree

accepted/0000-selective-versions-resolutions.md
```diff
@@ -0,0 +1,290 @@
...
+
+## Example
+
+For example with:
+```json
+ "devDependencies": {
+    "@angular/cli": "1.0.3"
+  },
+  "resolutions": {
+    "typescript": "2.3.2"
```
bestander: I still think it is a bit too brute force to force typescript to `2.3.2` everywhere.
We need to be able to enforce typescript to some subtree.
What do you think of the proposed glob for the resolution path?

The use case I am trying to solve is this:
- left-pad@2.0.5 is getting released with a bug
- it is used all over the app dependencies, some of them get locked on @1.0.1 which is fine 
- you want to upgrade jest to next version but some of the subdependencies requires "leftpad@^2.0.0" which will get resolved to "leftpad@2.0.5"
- you want to force jest's subdependency to leftpad@2.0.4 but don't change force it across the whole app

I would think this could solve it elegantly.
```
"resolutions": {
  "jest-*/**/left-pad": "1.0.4"
}
```

What do you think?

bestander:
@victornoel, this is one of the best RFC I've seen here, great job!

accepted/0000-selective-versions-resolutions.md
victornoel: yes, I agree, I was actually thinking of something like that after integrating in my head all the comments and @arcanis glob pattern proposition :)

I'm still processing all that was said and also things to add (check command in particular) to produce an update of the RFC.

victornoel:
I'm back on the idea you expressed @bestander:
> A maybe easier alternative is to override a specific semver with another one, e.g. typescript@>=2.5 < 3 => typescript@4.0.2

Do you mean by that that it would apply to all dependencies to typescript that comply with the version specification `>=2.5 < 3` or only to the dependencies exactly expressed with this semver?

Because I understood the second at first, but while rewriting the RFC I realised you maybe have meant the first (and it does make more sense actually, not sure it would be simpler to implement in that case though...).

bestander:
I actually thought of second as an easy way to identify a specific resolution that breaks my app.
But after thinking about it more I think an approach that takes path in consideration would be better

victornoel:
Ok thanks :) I am asking mostly to clearly explain the alternative propositions, which could be actually used as a way to extend the current proposition (the one with a glob pattern).

victornoel:
ok, here is an updated version with everything we discussed, I will be waiting for feedbacks from you both :)

Please take into account the questions at the end.

victornoel:
@bestander something is not clear to me with `check --verify-tree`: what is its usage?
If I understand correctly, it is meant to not use the yarn resolver, so for example, it does not use the resolutions when `--flat` is used, so would there be any reason to change its behaviour within the scope of this RFC?

As for `check` itself, I guess it should just check if the current lockfile and `node_modules` is coherent with whatever `install` would have done, as it already do currently, right? (i.e., there should be no change in abstract behaviour: within the scope of this RFC this means it takes resolutions into account of course).

WDYT?

accepted/0000-selective-versions-resolutions.md
```diff
@@ -0,0 +1,423 @@
...
+Most of the examples are given with exact dependencies, but note that using a
+non-exact specification in the `resolutions` field should be accepted and
+resolved by yarn like it usually does. This subject is discussed below also.
+
+Any potentially counter-intuitive situation will result in a warning being
+issued. This subject is discussed at the end of this section.
+
+## Examples
+
+We the following packages and their dependencies:
```
sejoker: the verb is missed, probably `have`?

accepted/0000-selective-versions-resolutions.md
```diff
@@ -0,0 +1,423 @@
...
+ "devDependencies": {
+    "package-a": "1.0.0",
+    "package-b": "1.0.0"
+  },
+  "resolutions": {
+    "package-a/package-d1": "3.0.0"
+  }
+```
+
+yarn will use `package-d1@3.0.0` oly for `package-a` and `package-b` will
```
sejoker: oly -> only

accepted/0000-selective-versions-resolutions.md
```diff
@@ -0,0 +1,423 @@
...
+- `**/a/b` denotes the directly nested dependency `b`
+of all the dependencies and nested dependencies `a` of the project.
+- `a` denotes the project's dependency `a` (see below for a discussion on
+the matter of specifying a *resolution* for one of the project dependency).
+- `a/**/b` denotes all the nested dependencies `b` of the project's
+dependency `a`.
+- `**/a` denotes all the nested dependencies `a` of the project.
+- `**/a-*` denotes all the nested dependencies of the project whose
+name starts with `a-`.
+- `**` denotes all the nested dependencies of the project (a bad idea mostly,
```
sejoker: I think we should limit usage of `**` and prohibit designations ending with `**` as I don't see the real use case for that scenario.

sejoker: so it will mean that only `**` will be also prohibited

accepted/0000-selective-versions-resolutions.md
```diff
@@ -0,0 +1,423 @@
...
+The `devDependencies` and `dependencies` fields always take precedence over the
+`resolutions` field: if the user defines explicitly a dependency there,
+it means that he wants that version, even if it's specified with a non-exact
+specification. So the `resolutions` field only applies to nested-dependencies.
+Nevertheless, in case of incompatibility between the specification of a
+non-nested dependency version and a *resolution*, a warning is issued.
+
+## Relation to the `--flat` option
+
+The `--flat` option becomes thus a way to populate the resolutions field for
```
sejoker: I would love to see an example when bot `resolutions` and `--flat` option are specified.

sejoker: I guess `--flat` implies it will override resolutions with single version.

accepted/0000-selective-versions-resolutions.md
```diff
@@ -0,0 +1,423 @@
...
+```
+
+With:
+```json
+ "devDependencies": {
+    "package-a": "1.0.0",
+    "package-b": "1.0.0"
+  },
+  "resolutions": {
+    "**/package-d1": "3.0.0"
```
sejoker: please specify that version is always in the form of 'major.minor.patch' and not allows any wildcards like `3.1.x` or `3.1.*`.

accepted/0000-selective-versions-resolutions.md
```diff
@@ -0,0 +1,423 @@
...
+  }
+```
+
+yarn would then replace matching version specifications with the user's one.
+For example a dependency normally resolved to `typescript@2.2.2` would be
+resolved in practice to `typescript@2.3.2`.
+
+## Mapping version specifications as well as packages name
+
+Same as the tow above but with a different name on the right-hand side of the
```
sejoker: typo -> `tow`

accepted/0000-selective-versions-resolutions.md
arcanis: From an implementation pov, I'd prefer not to add artificial limitations. We will never be able to completely prevent the users to shoot themselves in the foot with this kind of feature anyway, and we would have to pay the complexity toll.

accepted/0000-selective-versions-resolutions.md
victornoel: With this RFC, `--flat` becomes a flag that have an impact only on modifying the `package.json`, at the same level as `add`, `upgrade` or `upgrade-interactive` do.

It is not used anymore for `install` at all: `install` uses whatever is in the `resolutions` field (and the content of the later could have been populated manually or via the use of `--flat`).

But actually your question makes it clear that it could be useful to take `--flat` into accounts in checks too, to ensure that all dependencies of the project are actually flattened (via the `resolutions` field) as expected. I will add it to the RFC :)

accepted/0000-selective-versions-resolutions.md
victornoel: No they are not, non-exact versions are also allowed.

victornoel:
any feedbacks on my question about `check` above @bestander @arcanis @sejoker @xtuc? Before I push more changes to the RFC, I would like to include the `check` aspect of the question :)

bestander:
Thanks Victor.
This is on the top of my list.
Sorry if there is a delay, we need to push some urgent fixes.

victornoel:
@bestander ok, thanks for notifying me, this is no problem of course, I will wait :)

accepted/0000-selective-versions-resolutions.md
```diff
@@ -0,0 +1,423 @@
...
+
+Thus it would make sense to have a bit of the documentation talking about
+this use case and underlying the fact that *resolutions* are mostly here
+on a temporary basis.
+
+# Drawbacks
+
+## Teaching
+
+It makes yarn behaviour a bit more complex, even though more useful. So it
```
magicmark: nit: even though -> although

accepted/0000-selective-versions-resolutions.md
```diff
@@ -0,0 +1,423 @@
...
+going to be applied. In particular, I'm thinking about defining a *resolution*
+like that and latter adding a dependency which has a dependency that matches
+but is not intended to. I feel like it adds a lot of mental weight for the
+user to manipulate single `*`.
+
+## `--flat` option
+
+When used with the `install` command, the behaviour of `--flat` with respect
+to this RFC is clear: it will populate the `resolutions` for all packages with
+`**/dependency` as package designations.
```
magicmark: To clarify: is this proposing to add a resolution for _all_ `dependencies` in package.json to `resolutions`? Or just the ambiguous dependencies that yarn would currently prompt for anyway?

magicmark:
This is awesome! I really like the idea of storing the information in package.json's `resolutions`.

I would personally vote to alter the wording from 'nested' to 'transitive' (although fwiw npm uses ['secondary', 'nested'](https://docs.npmjs.com/how-npm-works/npm3)' and '[transitive](http://blog.npmjs.org/post/145724408060/dealing-with-problematic-dependencies-in-a)', so I'm possibly just bikeshedding :P)

How do you think this should relate to https://github.com/yarnpkg/rfcs/pull/54? It looks like the current interface to resolve nested dependencies and add them to `package.resolutions` in this RFC is through `yarn add|install --flat`, so maybe #54 is redundant? (Or does this overload the meaning of `--flat`?)

accepted/0000-selective-versions-resolutions.md
victornoel: Good question... I was going with `--flat` behaving as before, but writing the user choices in the resolutions field (so your second interpretation), but I'm not so sure it is the most obvious: the user expects all the tree to be flattened, the user doesn't care that today some dependencies resolve to only one version, so maybe the first interpretation would be better...

Actually, the more I think about it, and the more it feels like the `--flat` behaviour I'm advocating for is somewhat competing with the original behaviour of the lockfile (i.e., storing and locking resolutions at a given moment).

I'm unclear on what is best thus now... :)

victornoel:
@magicmark thanks for the feedbacks!

Concerning #54, I thought you wanted to be able to choose only some of the transitive dependencies, so `--flat` is maybe not the way to go, there seems to be something to do with #54, as a user of the present RFC I think :)
But I agree there is some kind of intersection between what `--flat` is doing and what your proposal is about.

I wonder if we shouldn't differentiate use cases (with some imaginative solutions to each, for the sake of discussion):
 - the user wants to manually override a version specification of a given transitive dependency because the original one is incorrect with respect to its needs: it thus overrides them via the `resolutions` field (manually or helped by a yarn command).
 - the user wants to apply the behaviour of `upgrade` but on a select number of transitive dependencies (note that `upgrade` only applies to transitive dependencies, while `upgrade-interactive` only applies to the project's dependencies!): the user would thus uses something like your proposal which could directly modifies the `yarn.lock`, no need to go through the `resolutions` field since it is not about override
 - the user wants to flatten the tree: in this case, it would make sense to talk about override, because some transitive dependencies WILL have their version overridden... again it's not clear which one should be stored, and also how the user can know if the resolutions became inconsistent (maybe with `check` as discussed above in another comment).

dfreeman:
> - the user wants to apply the behaviour of upgrade but on a select number of transitive dependencies (note that upgrade only applies to transitive dependencies, while upgrade-interactive only applies to the project's dependencies!): the user would thus uses something like your proposal which could directly modifies the yarn.lock, no need to go through the resolutions field since it is not about override

I'd agree that this is fundamentally a slightly different use case, which was why I liked #54 as a separate proposal. The user isn't forcing a version resolution, but something more akin to just doing a fresh install of the transitive dependencies, allowing them to resolve to a more recent (compatible) release on their own.

As you called out, it would be somewhat strange for this to wind up in `resolutions`, since the fact that they previously resolved to a different version is a product of the moment in time they were installed, not a hard constraint.

magicmark:
> the user would thus uses something like your proposal which could directly modifies the yarn.lock, no need to go through the resolutions field since it is not about override

So this _was_ my original plan. But one of the questions I had in my RFC was that if I'm overriding the version of a transitive dependency, it's probably for a good reason (security, critical bug fix etc). 

I therefore probably want this presumably crucial information to survive rebuilds of the lockfiles ([lockfiles can be destroyed and rebuilt](https://github.com/yarnpkg/yarn/pull/3544)).

This is why I said:
> I really like the idea of storing the information in package.json's resolutions.

because it addresses this issue. But we'd probably only want to do this for the specifically overridden or conflicting versions, to avoid replicating the whole lockfile in `package.resolutions`.

So here's the big question: should/could #54 use the extended logic of `package.resolutions` proposed here? (I'm personally tentatively leaning 'yes', but I'm very interested to hear what others think? maybe this overloads the behaviour of `package.resolutions` too much.)

The below shows how this _could_ (but not necessarily should) work 

- - - 

If we're both using `package.resolutions`, for basically the same thing, maybe all three use cases you listed can be solved this way.

Let's consider 3 uses cases that show how out solution could be generalized (let me know if there's any other scenarios here?)
1) I do `--flat` installs, and I want to store a conflict that has arisen
2) I do `--flat` installs, and I want to store an override to a non-conflicting dependency
3) I do non-flat installs, and I want to store an override to a non-conflicting dependency

**Scenario 1:**
> 1) I do `--flat` installs, and I want to store a conflict that has arisen

Yarn prompts for a decision to write to `package.resolutions`. It currently does so.

`package.json` looks like:
```
resolutions: {
  "foo": "1.2.3"
}
```

(note: this is a valid target in our glob syntax)

- In a flat install, foo gets resolved to 1.2.3
- If I now did a non-flat install in this application, if foo gets installed at the top level, yarn will now force it to version 1.2.3*. It will allow other versions of foo to be installed when nested.


**Scenario 2:**
> 2) I do `--flat` installs, and I want to store an override to a non-conflicting dependency

Yarn _does not_ prompt for a decision to write to `package.resolutions`. I either have to edit `package.resolutions` by hand, or use a new command like `upgrade-interactive`

`package.json` looks like:
```
resolutions: {
  "bar": "1.2.3"
}
```

or also valid:
```
resolutions: {
  "bar": "my-bar-fork@1.2.3"
}
```

- In a flat install, bar gets resolved to `my-bar-fork@1.2.3`
- If I now did a non-flat install in this application, if bar gets installed at the top level, yarn will now force it to version `my-bar-fork@1.2.3`. It will allow other versions of bar to be installed when nested.


**Scenario 3:**
> 3) I do non-flat installs, and I want to store an override to a non-conflicting dependency

Yarn _does not_ prompt for to write to `package.resolutions`. I either have to edit `package.resolutions` by hand, or use a new command like `upgrade-interactive`

`package.json` looks like:
```
resolutions: {
  "path/to/baz": "1.2.3"
}
```

- In a flat install, baz does not get overriden. The resolution path does not match ¯\_(ツ)_/¯
- In a non-flat install of this application, for the baz matching `path/to/baz`, baz is installed at 1.2.3

Or in the case of some security patch I maybe want to make sure **all** baz packages are always matched:
```
resolutions: {
  "**/baz": "1.2.3"
}
```

- In a flat install, `**/baz` will match the top level baz. it will be installed at 1.2.3
- In a non-flat install of this application, all baz packages are matched. They are all installed at 1.2.3 





\* caveat: this may alter what node_modules would otherwise currently look like. If it was the transitive version of foo at 1.2.3, and the top level version was at 2.0.0, we've now potentially introduced a breakage in an app across yarn versions. (Hopefully an application would not be relying on this contrived workflow, the 'fix' here would be to always do `--flat` or not). This should be discussed further, as it could be really confusing.

victornoel:
There is one problem in what you wrote @magicmark: the glob patterns of `package.resolutions` do NOT correspond to the file tree in `node_modules` but to the dependency tree!

This means `"bar: "1.2.3"` never matches anything, because `package.resolutions` only pertains to transitive dependencies. So I think your use case 2 and 3 are the same actually in terms of what is written in `package.resolutions`.
That's why I wanted to consider that `--flat` is about generating fields in `package.resolutions`, not about populating the `node_modules` directory (which will be still indirectly flattened when using `--flat` as a result of the `package.resolutions` generated by `--flat`). In other words, `--flat` is about choosing a set of dependency, not about generating `node_modules` like it currently does.

Also we were not planning to support `"bar": "my-bar-fork@1.2.3"` for now, but this is not a problem for the sake of discussion (and maybe actually I don't think it will be a difficult aspect to implement in this RFC, so maybe it will be supported, who knows).

@dfreeman good point, what is exactly, according to you, the problem with simply calling `upgrade` when there is the need to use the latest of the transitive dependencies? In other words, do you think the use case you selected really need #54?

magicmark:
So the glob matching should work when matching the dependency tree, and not the directory structure? (My wording above might need fixing)

It would be a proposed change to have yarn consider `package.resolutions` when doing a regular install, not just a `--flat` install.

In which case, why would `"bar: "1.2.3"` not match the top level dependency of bar?

dfreeman:
> good point, what is exactly, according to you, the problem with simply calling upgrade when there is the need to use the latest of the transitive dependencies? In other words, do you think the use case you selected really need #54?

@victornoel the specific use case I have in mind is something like this: I discover a bug in a transitive dependency of my project, open a PR to fix it, and ultimately there's a new patch release of that dependency. Upgrading the root dependency won't do anything, because it's already at the latest version, but I'd like to be able to refresh the version of that transitive dependency to get the fix.

Today I can do this by uninstalling and reinstalling the root dependencies that bring in the transitive one, but that can get messy fast if lots of my dependencies require that transitive one, and it also opts me into a bunch of other upgrades I may or may not want. I'd love to be able to say what I mean in this scenario and just do e.g. `yarn upgrade-transitive '**/some-package'`.

magicmark:
The above is also exactly what I'm after. So for me, the question boils down to:

1) should `yarn upgrade-transitive '**/some-package'` just install the update and save it in the lockfile? (the original behaviour of #54)
2) or should `yarn upgrade-transitive '**/some-package'` work by saving `"**/some-package": "x.x.x"` to `package.resolutions`, _then_ doing an install? (and have yarn understand it)

There's pros/cons to both, 1 is maybe simpler? Sorry to have slightly hijacked this thread!

victornoel:
> Today I can do this by uninstalling and reinstalling the root dependencies that bring in the transitive one, but that can get messy fast if lots of my dependencies require that transitive one, and it also opts me into a bunch of other upgrades I may or may not want. I'd love to be able to say what I mean in this scenario and just do e.g. yarn upgrade-transitive '**/some-package'.

Note that you can also do a `yarn upgrade` to achieve the same effect than removing and adding the root dependency (except it will apply to all transitive dependencies).

So, both @magicmark  and @dfreeman: I see the need yes, in that case, I think we agree that this use case is quite different from what is discussed in this RFC: as you rightly said, this means there is some need for what is described in #54, i.e. a way to select which package to upgrade with `yarn upgrade`.
So something like `yarn upgrade '**/some-package'` would make sense to me, to complete your proposition @magicmark. This could be discussed in #54 thus :)

About naming: #54 proposes applying `upgrade-interactive` to transitive dependencies, but I wouldn't do that personally, because `upgrade-interactive` is already used to modify the `package.json` file, so it could be really confusing in term of usage :).

Actually, there was good reason to discuss that here (so no problem with the hijacking ;), because we could consider that doing: `yarn upgrade-interactive '**/some-package'` would allow to generate valid `package.resolutions` from the command line! So I will maybe include it in the RFC too, at least for discussing it!

@magicmark about your question on `"bar": "123"`, this is explained in the RFC actually, but in a few words:
 - it matches only a root dependency of the project (either something in `package.dependency`, `package.devDependencies` or `package.optionalDependencies`), nothing else. Even with `--flat`, it is not because `bar@123` is present at the root of the `node_modules` that it is not a transitive dependency anymore. `--flat` is about resolutions, not about the dependency tree (even though we say "flattening the dependency tree", in practice each dependency has its own dependencies, and they form a tree, the only thing that changes with `--flat` is that it is the same version used everywhere, so yarn can optimize the `node_modules` by putting all dependencies at the root of the directory.
 - it is ignored by the solution proposed in this RFC because we can consider that if the user wants to force a given direct dependency, then he can do it in `package.dependency`, `package.devDependencies` or `package.optionalDependencies` (and it should raise a warning saying the `resolution` is not used).

dfreeman:
> About naming: #54 proposes applying upgrade-interactive to transitive dependencies, but I wouldn't do that personally, because upgrade-interactive is already used to modify the package.json file, so it could be really confusing in term of usage :).

Just a note for clarity: the title on that RFC still mentioneds `upgrade-interactive`, but I think the actual content has shifted to suggest introducing a new `update-transitive` command (though I personally like `upgrade-transitive` for consistency 🙂 )

sejoker:
@magicmark sure, let me read the latest discussion and put my thoughts on the topic into order.

bestander:
@victornoel, how close do you think we are to get this merged?
I don't want to drag this spec for too long so that you don't loose interest in it :)
This feature is very important, we definitely should have it in

victornoel:
@bestander actually I'm waiting from feedbacks from you about some questions I had (https://github.com/yarnpkg/rfcs/pull/68#issuecomment-304456048) and maybe some comment on what has been said after that :)

bestander:
>  something is not clear to me with check --verify-tree: what is its usage?
If I understand correctly, it is meant to not use the yarn resolver, so for example, it does not use the resolutions when --flat is used, so would there be any reason to change its behaviour within the scope of this RFC?

--verify-tree was built to make sure that all packages inside node_modules are consistent between each other independently of Yarn's resolution logic.
For now we probably don't need to make changes to it. 
If you force a version that does not match semver requirements of a packagem `--verify-tree` would throw an error.
We can expand --verify-tree later to support the overrides too but I would not bother now.

> As for check itself, I guess it should just check if the current lockfile and node_modules is coherent with whatever install would have done, as it already do currently, right? (i.e., there should be no change in abstract behaviour: within the scope of this RFC this means it takes resolutions into account of course).

The default check reads yarn.lock and makes sure that all versions in it match to what is inside node_modules.
I think you should get this for free without extra changes

accepted/0000-selective-versions-resolutions.md
```diff
@@ -0,0 +1,423 @@
...
+
+But when used with the `add` command, should it apply only to the nested
```
bestander: Is it a specific case of add command conflicting with resolutions?

E.g. `yarn add left-pad@1.0.0` could conflict with `resolutions: {"left-pad": "1.3.0"}`

bestander:
This looks good to me if there are no more open questions and if all typos are fixed.

I think it is quite doable, all the resolutions are done recursively https://github.com/yarnpkg/yarn/blob/master/src/package-resolver.js#L425 but they keep parent references https://github.com/yarnpkg/yarn/blob/master/src/package-request.js#L308, so matching the resolutions pattern could be done there.

victornoel:
@bestander ok, I will update the RFC this week-end and we should be good to merge :)

accepted/0000-selective-versions-resolutions.md
victornoel: no, it is about what would be the behaviour of `yarn add --flat left-pad`:
1. should it flatten (i.e. write *resolutions* for) only the dependencies of `left-pad` without flattening the rest of the nested dependencies of the project (except for those that are common to `left-pad`)
2. should it flatten (i.e., write *resolutions* for) the whole project (as it currently does).

bestander:
Ah.
From Yarn perspective--flat is the node in which package installation runs.
So it should flat the whole project.

victornoel:
> From Yarn perspective--flat is the node in which package installation runs.  So it should flat the whole project.

I'm not sure I understand what you meant by that :) did you mean "mode"?
The thing is that I propose to change the meaning and behaviour of `--flat`, so maybe it shouldn't be the same? anyway, maybe as a starter we don't change that behaviour, and `--flat` stays the installation mode, so it impacts the whole project dependency tree and ensure at all time that things are flattened (as it already does).
The only different from before is that once some *resolutions* are stored in the package.json, they are used as-is even if `--flat` is not specified.
Does it seem good?

bestander:
> I'm not sure I understand what you meant by that :) did you mean "mode"?

Typo, edited to clarify.

> The thing is that I propose to change the meaning and behaviour of --flat, so maybe it shouldn't be the same? anyway, maybe as a starter we don't change that behaviour, and --flat stays the installation mode, so it impacts the whole project dependency tree and ensure at all time that things are flattened (as it already does).
The only different from before is that once some resolutions are stored in the package.json, they are used as-is even if --flat is not specified.

For now I propose to keep it the same.
We can discuss other options later on if there is a need.

victornoel:
I have pushed another update.

In particular I have changed something related to retro-compatibility (in a few words, the package designation `package-a` would be an alias for `**/package-a`, see the RFC for rational).

I'm sorry but there is still some open questions:
 - about using single `*` in package designations (not very important question)
 - about `resolutions` specified in nested dependencies `package.json` (this must be clarified, but the answer seems straightforward, i.e., do not consider them)
 - still about `--flat` and what it means, because I really feel like the current choice of keeping it as-is is a mistake in terms of clarity of its semantics :) (this must be clarified, I do think something must be done, maybe later though).

There is also the need to review maybe again all of it, if there is people up to the task, to be sure no contradictions crept in with all the changes :)

accepted/0000-selective-versions-resolutions.md
```diff
@@ -0,0 +1,495 @@
...
+
+The two alternatives discussed in the section just above, "Mapping version
+specifications" and "Mapping version specifications as well as packages name",
+can be adapted to the current proposition to support these uses cases as well.
+
+# Unresolved questions
+
+## Package designation and single `*`
+
+I am not totally convinced that using a single `*` in a package designation
```
bestander: Ok, let's not do single *

accepted/0000-selective-versions-resolutions.md
```diff
@@ -0,0 +1,495 @@
...
+like that and latter adding a dependency which has a dependency that matches
+but is not intended to. I feel like it adds a lot of mental weight for the
+user to manipulate single `*`.
+
+## `--flat` option
+
+I wonder if the `--flat` option of `install` shouldn't be transformed to
+a `flatten` command that would:
+1. Fill in the *resolutions* for all nested dependencies.
+2. Set the `flat` field in the `package.json`.
```
bestander: I like that idea.
Unlike --production, --flat is an option that you decide for the project early on and stick to it.

accepted/0000-selective-versions-resolutions.md
```diff
@@ -0,0 +1,495 @@
...
+
+If we do that, then the `--flat` option of `add` (and the `flat` option in the
+`package.json`) would apply not to the installation but to the adding,
+upgrading, etc (everything that modify the `package.json`'s dependencies):
+it will ensure that the project stays flattened via the populating of the
+`resolutions` field.
+
+## resolutions in dependencies
+
+If a dependency contains resolutions, should it be taken into account?
```
bestander: No, same as shrikwraps/lockfiles, every project is responsible for their resolutions.
Also right now this is a Yarn-only feature, we don't really want to get it into the wild yet too much

accepted/0000-selective-versions-resolutions.md
```diff
@@ -0,0 +1,495 @@
...
+a `flatten` command that would:
+1. Fill in the *resolutions* for all nested dependencies.
+2. Set the `flat` field in the `package.json`.
+
+It makes no real sense to have a flattening mode for `install`:
+1. `install` already follows the `resolutions` field with this RFC.
+2. `install` should be only about building the `node_modules` directory, not
+modifying the the `package.json` IMHO.
+
+If we do that, then the `--flat` option of `add` (and the `flat` option in the
```
bestander: I don't understand how much --flat affects resolutions feature.
Can we decide on flat in a follow up RFC?

bestander: If --flat gives trouble to this feature then let's just throw on `--flat` if resolutions property is specified

accepted/0000-selective-versions-resolutions.md
victornoel: I agree :)

accepted/0000-selective-versions-resolutions.md
victornoel: ok, cool!

accepted/0000-selective-versions-resolutions.md
victornoel: I'm not sure I understand what you mean:
1. Currently, `--flat` is both about populating `resolutions` field AND taking `resolutions` field into account when doing `install` (including install after `add`).
2. This RFC is about taking `resolutions` field into account when doing `install` (including install after `add`).
3. So with this RFC, `--flat` is only about populating `resolutions` field.

Thus the question is when should the `resolutions` field be populated.
Maybe this can simply be managed in another RFC that transforms `--flat` into:
 - a `flatten` command that flattens the current dependency tree.
 - a `--flat` option (for all `package.json` modifying command such as `add`) as well `flat` field to keep the dependency tree flattened at all time.

bestander: I agree, it seems like I might not have all the context on this one.
I thought that --flat was just enforcing that every dependency is resolved only once so that after hoisting we get a flat structure in node_modules.

I propose leaving --flat behavior out of scope for this RFC and move forward with the implementation if it is possible.
We can follow up with --flat specific right after it.

accepted/0000-selective-versions-resolutions.md
```diff
@@ -0,0 +1,495 @@
...
+    "package-a": "1.0.0",
+    "package-c": "1.0.0"
+  },
+  "resolutions": {
+    "**/package-a/package-d1": "3.0.0"
+  }
+```
+
+yarn will use `package-d1@3.0.0` both for `package-a` and the nested
+dependency `package-a` of `package-c`.
```
arcanis: Why? `**/package-a/package-d1` doesn't match `package-a`

arcanis: Unless you mean that both `/package-a/package-d1` and `/package-c/package-a/package-d1` will be fixed at `3.0.0`?

accepted/0000-selective-versions-resolutions.md
```diff
@@ -0,0 +1,495 @@
...
+```json
+  "resolutions": {
+    "package-a": "1.0.0"
+  }
+```
+
+With the current proposal, the package designation `package-a` is an alias for
+`**/package-a`: this means the behaviour of yarn with a project whose
+`resolutions` field contains *resolutions* filed by a pre-RFC yarn will be
+as expected: the nested dependencies will have the fixed version specified.
```
arcanis: Good idea. I also think we should prevent people from using both `pkg-name` and glob syntaxes at the same time. If they want the same effect, they should explicitely use `**/pkg-name`.

arcanis:
This RFC seems to be in a really good shape @victornoel, kudos! 😃 What would you think about implementing a first iteration in Yarn and see how it goes? Would you be willing to do it?

accepted/0000-selective-versions-resolutions.md
victornoel: > Unless you mean that both `/package-a/package-d1` and `/package-c/package-a/package-d1` will be fixed at `3.0.0`?

yes: `/package-a/package-d1` and `/package-c/package-a/package-d1` both match `**/package-a/package-d1`

accepted/0000-selective-versions-resolutions.md
victornoel: Do you mean:
1. consider both syntaxes equally valid but prevent the user from mixing them in the same project, or
2. have a preference for the glob syntax and keep the non-glob syntax only for retro-compatibility purposes?

How would we achieve one or the other btw? Via printing a warning? For 2. I can easily see it happening, but for 1. it would seem strange I think...

victornoel:
@arcanis last time I checked the code for this task, I got a bit lost in it, I'm worried I won't succeed doing it and I wouldn't want to commit to something I can't do :)
@bestander seemed to know which specific places would be concerned (see https://github.com/yarnpkg/rfcs/pull/68#issuecomment-308897162), so, once we consider the set of features to implement fixed, I can look at the code again with @bestander guidance on where to look (I suppose this won't be different from what he envisioned in the above comment) and tell you if I can do it or not :)

I'm going to introduce the changes discussed in last few days very soon and then I will explore the code again!

bestander:
@victornoel, you'll have our full support.
Feel free to push some half-baked ideas in and we can have a look.
Maybe you could start with a PR that adds failing and skipped tests that cover the spec?
Then with another go we would just need to comply with the tests

bestander:
Personally, I think this is one of the most important feature in while for Yarn.
It addresses a huge point of stress for semver packages users.

victornoel:
> Maybe you could start with a PR that adds failing and skipped tests that cover the spec?

Actually I've only started doing that some time ago :) it's a good idea, and like this, worst case scenario is someone will be able to implement the RFC from the failing tests :)

victornoel:
ok, I pushed the last changes, I kept the discussion on the future of `--flat` in a specific section at the end for record, and later we should make another RFC.

bestander:
Great job, @victornoel.
This is going to be huge!

s-panferov:
This RFC doesn't play well with scoped packages, because it uses `/` as a path separator. Please see https://github.com/yarnpkg/yarn/issues/4242

HaleTom:
Where do I find the official location of this RFC (and its number)?

bestander:
@HaleTom, it lives in this repo after it is merged https://github.com/yarnpkg/rfcs/blob/master/accepted/0000-selective-versions-resolutions.md

victornoel:
@bestander by the way, maybe it should be moved to the implemented sub-folder now, no?

bestander:
Of course, do you want to send a quick PR?

victornoel:
ok, I will do that in the next few days! :)

PR #69, opened 2017-06-14
Soft Engines Check

gsklee:
See [here](https://github.com/gsklee/rfcs/blob/master/text/0000-soft-engines-check.md).

### DIFF
--- a/text/0000-soft-engines-check.md
+++ b/text/0000-soft-engines-check.md
@@ -1,0 +1,87 @@
+- Start Date: 2017-06-14
+- RFC PR: (leave this empty)
+- Yarn Issue: (leave this empty)
+
+# Summary
+
+Make the `engines` field in `package.json` produces soft warnings instead of hard
+errors during `yarn install`/`yarn`.
+
+# Motivation
+
+See [#3430](https://github.com/yarnpkg/yarn/issues/3430#issuecomment-303392013).
+`npm` has deprecated the `engineStrict` field in `package.json` since v3, and all
+`engines` setting violations will only produce warnings
+<sup>[[1]](https://github.com/npm/npm/releases/tag/v3.0.0)</sup>
+<sup>[[2]](https://docs.npmjs.com/files/package.json#engines)</sup>; Yarn, on the
+other hand, still throw hard errors:
+
+```shell
+error some-package@3.1.13: The engine "node" is incompatible with this module. Expected version "v7.6.0".
+error Found incompatible module
+```
+
+This has caused considerable inconveniences due to several reasons:
+
+* People download npm packages to include part or whole of them in client-side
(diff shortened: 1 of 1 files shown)


ljharb (approved):
This is the proper default; thanks!

kittens:
We specifically chose a strict engine check with Yarn to encourage the ecosystem to properly use the `engines` field. Relaxing this would make the `engines` field useless and encourage already rampant misuse of it.

This justification in this RFC for relaxing this is mostly "npm doesn't do this and have deprecated it". Obviously we aren't adherent to the changes of npm and we deliberately deviated.

gsklee:
I'm convinced. Thanks for the inputs guys 👍

bestander:
Thanks, @gsklee, for leading the conversation and cleaning up the tasks, much appreciated.

hbetts:
Could this behavior be communicated to the community via Yarn's blog? (Similar to the excellent blog post on the differences in the deterministic behavior of each package manager)

Also, if it's alright with the core team, could @gsklee submit an RFC that documents this behavior? I'd like to be able ot point others to a spec.

bestander:
+1 on documentation or a blog post.
Anyone can to send PRs with blog posts or documentation changes, they are always welcome.

gsklee:
I can write a post on this if @kittens or @samccone are not able to find the time to do it.

bestander:
Go ahead with the post, G. Kay, of you have time!

havenchyk:
sorry for asking here, but is there a way to check the yarn version that the team members use for development (and fail on e.g. `yarn start`)? I thought `engine` will help, but it didn't

samccone:
@havenchyk this seems like an interesting feature, mind creating another rfc?

rattrayalex-stripe:
@havenchyk @samccone was another RFC created? can it be linked here? I too would like that feature

havenchyk:
@rattrayalex-stripe nope, it wasn't. Still on my todo list, but if you would like, take care of it

BorntraegerMarc:
Could yarn at least log which module is causing the error? I'm just getting the error
```
error @angular/cli@1.2.4: The engine "node" is incompatible with this module. Expected version ">= 6.9.0".
error Found incompatible module
```
in my package.json I have exactly:
```
    "engines": {
        "node": ">= 6.9.0",
        "npm": ">= 3.0.0"
    },
```

So I don't know who/what is causing the error...

hbetts:
@BorntraegerMarc please run `node --version` to see which Node version you have installed.

BorntraegerMarc:
@hbetts ah that explains it. Version `v6.1.0` was installed. OK maybe just a better error message if the actual host is mismatching 😄

MrCoder:
Can we include a message like "use --ignore-engines to bypass engines check"?

PR #70, opened 2017-06-18
add RFC for script run hooks

deecewan:
This was spurred by https://github.com/yarnpkg/yarn/issues/3209#issuecomment-308606581 to try and normalise the hooks that run and when.

### DIFF
--- a/text/0000-update-hook-runs.md
+++ b/text/0000-update-hook-runs.md
@@ -1,0 +1,196 @@
+- Start Date: 2017-06-19
+- RFC PR:
+- Yarn Issue:
+
+# Summary
+
+Bring the scripts that run on certain commands closer to parity with NPM.
+
+# Motivation
+
+There are a few changes to the scripts hooks that NPM has made recently, and it
+would be good to bring `yarn` to the same standard.
+
+This means that `yarn` can remove hooks that don't make sense, and stop
+supporting issues brought up by people confused about *why* a certain script is
+running.
+
+Additionally, by tracking closely to what NPM is doing with their hooks, `yarn`
+remains a drop-in replacement, as most people expect it to act.
+
+# Detailed design
+
+### State of Scripts
+
+Below are all the scripts that NPM currently auto-runs, and at which points they
+run. Currently, there is only one deprecation that motivated this RFC, but it is
+anticipated that there will be more in the long run.
+
+This list is kind of long, because I've enumerated all options.
+
+- `prepublish`
+  - before `publish`
+  - before `install`
+    - *Note: this is the deprecation. `prepublish` will no longer run on
+      install*
+    - This hook should *not* be added, as it is removed as of `npm@5`.
+- `prepare`
+  - before `publish`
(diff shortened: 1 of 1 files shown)


text/0000-update-hook-runs.md
```diff
@@ -0,0 +1,196 @@
...
+  - runs *before* the version commit is made
+- `postversion`
+  - after `version`
+  - runs *after* the version commit is made
+- `preshrinkwrap`
+  - before `shrinkwrap`
+- `shrinkwrap`
+  - before `shrinkwrap`
+    - confusing, but runs *before* the shrinkwrap is created
+  - I'm not sure `yarn` needs to include this, given that `shrinkwrap` is not
```
bestander: yeah, we don't need shrinkwraps.

bestander:
Great work, @deecewan!

bestander:
Let me summon a few folks who have more context

bestander:
My opinion - this looks fine, we should just match npm@5 and move forward.
Are there any tricks we need to address here?
I remember seeing a few issues where people are expecting to run post/preinstall scripts when reinstalling node_modules.

deecewan:
AFAIK, from the docs and from testing, I've enumerated how everything works. I might make a couple of test scripts for `npm@5` to 'codify' the expected behaviours.

bestander:
Great job, @deecewan!
Let's do it! Are you up to send a few PRs?
Tests would be very important to make sure we don't regress

deecewan:
yeah, keen!

I'll do what I can 👍

andreineculau:
FYI a programmatic comparison of which hooks run and in what order, along with the toold can now be found at https://github.com/andreineculau/package-json-scripts

bestander:
@deecewan, any chance you've had a crack at this?
@andreineculau, do you want to help out?

text/0000-update-hook-runs.md
andreineculau: the shrinkwrap scripts also run when creating the new `package-lock.json`, so yarn should also run if when creating the `yarn.lock` file, for a closer 1-to-1 match

andreineculau:
@bestander I'm both limited by time and incentive as I do not use yarn in my routine. I've done the package-json-scripts investigation because we intend to have git dependencies, and I've seen inconsistencies. My first npm alternative to try out is pnpm actually, so trying out and even committing to yarn is too much in the future.

What I can help out with, no doubt, is doing one final manual testing of whatever new behaviour lands or is about to land in yarn, if nobody wants to run https://github.com/andreineculau/package-json-scripts themselves.

deecewan:
@bestander i started taking a look at this. `yarn` already does 95% of theses hooks in the correct order. I went through and figured out which hooks run and when, and then went through yarn and checked on them.

from memory, i think there were only 1 or 2 changes. i'll take another look after work tonight.

bestander:
@deecewan, awesome, thank you very much.
Do you think you could set up a couple of tests that cover this contract?

PR #73, opened 2017-07-11
Update knit proposal

devongovett:
Based on my experiments on https://github.com/yarnpkg/yarn/pull/3753, I'm updating the knit RFC with a few additional sections as requested by @bestander:

1. Added `yarn install --knit` option to automatically link dependencies when installing.
2. Section comparing several other yarn features for linking deps.
3. Mentioned requirement of `--preserve-symlinks` option in node for this to work correctly.


bestander:
Awesome!
Thanks, @devongovett, for stepping up

accepted/0000-yarn-knit.md
```diff
@@ -52,6 +52,24 @@ This behaves like `yarn add dep` except that it looks at the versions of `dep` t
...
+
+* `link://` dependencies are similar to `yarn link` but actually closer to what `yarn knit` would provide. `link://` dependencies also cause the linked module's dependencies to be installed locally, so as long as there is no `node_modules` folder inside the linked dependency, it would work identically to `yarn knit dep`. The existing implementation of `link://` dependencies could be used to implement `yarn knit dep`.
+
+* `file://` dependencies cause files to be copied instead of linked. This is not as useful because it means you must re-install every time a change is made to the dependency.
+
+* `yarn pack` + `yarn install dep.tgz` is similar to `file://` dependencies. The pack + install process must be re-run every time a change is made. It does correctly dedupe dependencies, however, as `node_modules` are excluded by `yarn pack`.
+
+* [Workspaces](https://github.com/yarnpkg/rfcs/pull/66) are similar but solve a different problem. Where workspaces are great for a tree of related modules (e.g. a monorepo), `yarn knit` is for linking together modules in separate trees, e.g. things that might be shared between multiple workspaces.
+
```
bestander: Nice overview.

accepted/0000-yarn-knit.md
```diff
@@ -60,18 +78,16 @@ This proposal is mostly additive and affects only how people work on libraries t
...
+Finally, another issue is with the way the node require resolution algorithm works. Dependencies of symlinked modules are resolved relative to the realpath, not the symlink. This means that you'll still get duplicates if both modules depend on a third dependency, or errors if that dependency is not installed in either place. This is solved by the recently added runtime option for node `--preserve-symlinks`, which skips getting the realpath when resolving modules. Something similar would need to be added to browserify/webpack to solve this there as well. I recently opened a [PR for browserify](https://github.com/substack/node-browserify/pull/1742) to support the same option.
```
bestander: I am not sure this is a problem with knitting because on install the knitted dependencies should not have subdependencies packed in the "repository".
So all subdependencies should be installed fresh at referer level.

bestander:
@devongovett, what do you think of making `knit` just a for of `link` and not a separate command?
E.g. all `link:/` dependencies would be knit instead of symlinked to the source if yarn `install/add/upgrade` is called with `--knit-link` flag?

That might reduce the terms overload on users.

accepted/0000-yarn-knit.md
devongovett: The issue is that the linked module may still have a `node_modules` folder where you are developing. When you `require('dep')` from your app, it gets resolved using the realpath, e.g. `~/dev/dep/index.js`. If that file has `require('dep2')`, it is resolved to `~/dev/dep/node_modules/dep2`, and that copy will be used instead of the one installed in your app. I verified this to be the case. It is not the case with the `--preserve-symlinks` option, because the resolution algorithm uses the symlink path (`~/dev/app/node_modules/dep/index.js`) instead of the realpath, so when resolving `dep2`, `~/dev/app/node_modules/dep2` will be used instead.

bestander: Does knitting mean that dependencies need run prepublish script and the files listed in `package.json/files` should be copied to some temp location?

I think there would not be a symlink `~/dev/app -> ~/dev/dep`, it would probably be `~/dev/app -> (yarn-cache)/dep-knit-<uuid>` and node_modules would be empty there.

devongovett: Copying files kinda defeats the purpose, no? Then you'd have to run `yarn knit` every time you make a change to a dependency. I think the symlink would be `~/dev/app/index.js -> ~/.yarn-knit/dep/index.js -> ~/dev/dep/index.js` for example. The realpath is still the last one, which may have a `node_modules` folder in it.

bestander: Then there is a contradiction we need to sort out:

https://github.com/devongovett/rfcs/blob/50712e27d2d242bf466224186fcbb87b9bbab0ba/accepted/0000-yarn-knit.md#isolating-node_modules-correctly

https://github.com/devongovett/rfcs/blob/50712e27d2d242bf466224186fcbb87b9bbab0ba/accepted/0000-yarn-knit.md#faithfully-representing-the-node_modules-hierarchy

https://github.com/devongovett/rfcs/blob/50712e27d2d242bf466224186fcbb87b9bbab0ba/accepted/0000-yarn-knit.md#a-practical-proposal----knitting

The links above indicate that knitting would simulate full npm publishing but locally.
Symlinking `app -> dep` breaks this simulation because `dep` may will have non publishable artifacts (node_modules, .yarnignore files etc)

devongovett: That's what the middle symlink is for. Only the published files would have a symlink in `~/.yarn-knit/dep`, so (as long as `--preserve-symlinks` is used) non-published files would not show up to the app.

bestander: I see.
Then node_modules would not be symlinked and this is not a problem.

devongovett:
@bestander so you mean the following?

* `yarn link --knit` - same as `yarn knit` in RFC
* `yarn add dep --knit` - does what `yarn knit dep` did in the RFC
* `yarn install --knit` - links all dependencies that have been setup with `yarn link --knit`

bestander:
* `yarn link --knit` - same as `yarn knit` in RFC

Yes

* `yarn add dep --knit` - does what `yarn knit dep` did in the RFC

Only if `dep` is like `dep@link:../path/to/dep`

* `yarn install --knit` - links all dependencies that have been setup with `yarn link --knit`

And also all dependencies with `link:/` and `file:/` (when used as link:/) specifier.

devongovett:
1. is fine
2. doesn't make sense to me because you'd need to know the path to the global yarn registry of knit dependencies to specify the link path. I'm on board with `yarn add dep --knit` means install `dep` from the knit registry though.
3. So it would treat `file://` deps as `link://`? I suppose so, but this seems separate from knit - i.e. it has nothing to do with deps in the knit registry.

devongovett:
cc @ide who wrote the original knit proposal

bestander:
What I tried to say that `knit` is a mode of `link` feature, the implementation could probably piggy-back ride on the link with a conditional action at the end.

And we can do linking 3 different ways:
- yarn link command
- yarn install/add/upgrade with package.json that contains link: dependencies
- yarn install/add/upgrade with package.json that contains file: dependencies but called with `yarn-link-file-dependencies true` in yarnrc, this basically makes them behave as link: dependencies

devongovett:
Ah I see, sorry for the misunderstanding.

Yes, I think knit could be a mode of link:

* `yarn link --knit` - same as `yarn knit` in RFC.
* `yarn link dep --knit` - same as `yarn link` now, but also installs dependencies of `dep` locally.

What do you think of dropping the need for `yarn link dep` and just using `yarn add dep --knit` instead though? This would have the effect of also automatically linking any sub-dependencies of `dep` which might have been setup locally (with `yarn link --knit`) as well, since implementation of `add` is shared with `install`. This might have happened previously with `yarn link` because it included `node_modules`, so sub-dependencies of `dep` that were linked would also be carried over.

bestander:
> Yes, I think knit could be a mode of link:

> * yarn link --knit - same as yarn knit in RFC.

👍 
What do you think of giving it a more self explanatory name like `--prepublish` or `--prepare`?

> * yarn link dep --knit - same as yarn link now, but also installs dependencies of dep locally.

Considering that at step one, `yarn link --knit`, we simulated a publishing to npm by putting files to a temp/cache folder it should create a symlink not to the folder of knitted project but to the folder where we put the build artifacts.
So it would work slightly differently.

> What do you think of dropping the need for yarn link dep and just using yarn add dep --knit instead though?

I am worried that it will be confusing.
Would it create a symlink or would it install `dep` by copying files?
`link` implies that the dependency will be installed via a link from a local place.

devongovett:
Sorry for the delay in reply.

Yes, I think the `knit` name isn't very descriptive. In my POC pull (https://github.com/yarnpkg/yarn/pull/3753), I called it `yarn link --deep` because it makes links to each file rather than the directory. Maybe that makes sense? Ideally this would be the default mode for `yarn link`, but I suppose we can't change compatibility with `npm link`. Maybe a config option in `yarnrc` to make it the default?

`yarn add dep --knit` is a bit confusing, but what about `yarn add dep --link`? Then it's clear it's linking. It also seems clearer than `yarn link dep` since it is both linking and installing dependencies, whereas `yarn link dep` is only linking.

So my proposal would be:

* `yarn link --deep` - links each file in the cwd to the global link registry
* `yarn add dep --link` - links dep to the global link registry and installs its dependencies
* `yarn install --link` - installs all dependencies and auto-links ones that are in the global link registry

No new commands, only two new options.

ide:
That naming proposal mostly makes sense to me. I prefer a flag other than `--deep` like `--pack` (as in `yarn pack`) or `--publish` since part of the RFC's intent is to mimic what happens when you publish a package. If your package has `src` and `build` directories and your package.json specifies `"files": ["build"]`, then only the `build` directory should get symlinked (and shallowly, not the files within it).

I like the `link://` URLs too, really useful for internal projects.

devongovett:
I think `yarn link --pack` is probably the best option so far. `--publish` might be confused with actual publishing I guess.

devongovett:
Another open question is what put in the lock file. I would say that installing with the `--link` option would not affect the lock file: it would remain as if you had installed from the normal package registry. Otherwise we might run the risk of committing changes to a lock file that don't work in a real environment. The only case where this doesn't work is if the package isn't published to the registry yet. In that case I guess it would work as `link://` urls do now, and you would have to update it manually when the package is published. Thoughts?

bestander:
`link --pack` sounds good

bestander:
As for yarn.lock - good question.
What would be the result of this operation:
```
yarn link project-b --pack
yarn add left-pad
```

Would the second command overwrite what is inside node_modules?
I think lockfile should indicate something to conform with the promise that the same yarn.lock produces the same node_modules

devongovett:
yeah its a bit weird if you commit your `yarn.lock` file though, since you may end up with local-only artifacts being committed.

devongovett:
Updated the RFC with the new naming as discussed above, and renamed the file. Sorry if it makes it harder to see the diff. :/

devongovett:
@bestander @ide thoughts on the update? Would like to get started on an implementation if you think this is good.

accepted/0000-link-options.md
```diff
@@ -0,0 +1,93 @@
...
+
+## A practical proposal
+
+This is a proposal that solves all of the problems above and isn't too hard to implement or understand. Conceptually, we find all the files we'd normally publish to npm, pack them up using symlinks instead of copies of the files, publish the pack to a local registry (just a directory), and then when installing we look up packages in the local registry directory instead of npm.
+
+### Running "yarn link --pack" inside of `dep`
+
+This is the step that simulates publishing `dep`. Running `yarn link --pack` in `dep` finds all the files that "yarn publish" would pack up and upload to npm. Crucially, this excludes `node_modules`, and would follow the same algorithm as "yarn publish" such as reading package.json's `files` field.
+
+Then it simulates publishing `dep`: it creates a directory named `dep-X.Y.Z` (where `X.Y.Z` is the version of `dep` in its package.json) inside of a global directory like `~/.yarn-link`. A symlink is created for each file or directory that `yarn publish` would normally have packed up. This step shares some conceptual similarities with publishing to a registry, except it uses symlinks and it's local on your computer.
```
bestander: Should we use cache folder for this instead?
If we opt in to a new ~/.yarn-link it will fall out of OS cache radars and becomes a risk of growing out of control.

How about `<yarn cache>/links/`?

accepted/0000-link-options.md
```diff
@@ -0,0 +1,93 @@
...
+
+## A practical proposal
+
+This is a proposal that solves all of the problems above and isn't too hard to implement or understand. Conceptually, we find all the files we'd normally publish to npm, pack them up using symlinks instead of copies of the files, publish the pack to a local registry (just a directory), and then when installing we look up packages in the local registry directory instead of npm.
+
+### Running "yarn link --pack" inside of `dep`
+
+This is the step that simulates publishing `dep`. Running `yarn link --pack` in `dep` finds all the files that "yarn publish" would pack up and upload to npm. Crucially, this excludes `node_modules`, and would follow the same algorithm as "yarn publish" such as reading package.json's `files` field.
```
bestander: The disadvantage is that you need to run `yarn link --pack` every time you want to iterate.
What do you think of running the `--pack` at the time of link consumption?

1. `yarn link` would register `dep` as "linkable"
2. `yarn link dep --pack` would run the packing and installation

bestander:
@devongovett, I think we are almost there, thank you for your work on this feature, it should be really useful.
I've added a few suggestions above

accepted/0000-link-options.md
devongovett: Potentially, though there should only be symlinks in the global registry anyway, so I don't think it should get that large. Putting it in the cache folder might cause your linked modules to suddenly stop working e.g. after a reboot, which would be weird IMO.

accepted/0000-link-options.md
devongovett: Not quite. You have to run `yarn link --pack` if once to set up, and only again if you add/remove a file in the root. If you add a file to a subdirectory (e.g. `src/`) it should automatically show up everywhere since we symlink directories in the root. I think this is reasonable. I definitely don't want to have to run a command every time any change is made: might as well use `yarn pack` + `yarn install` in that case.

bestander: Ah, I've missed that point.

I understand that your goal for this change is to improve iteration speed when you are actively developing `dep` that is used in one of your projects.

I think the goal should be to achieve correctness of `dep` and simulate full publish/install cycle.
We could think of performance tweaks in the next iteration and for people who absolutely need speed will just set up their environment to use vanilla `yarn link`.

Am I right you want to have this setup?

```
app/node_modules/dep ->(symlink) ~/.yarn-links/dex-x.y.z ->(multiple symlinks) dep
```

I think you won't be able to achieve the right isolation with symlinks inside `~/.yarn-links/dep-x.y.z -> dep`. You'll end up cherry-picking symlinks of files and folders while respecting rules in .yarnignore/.npmingore.

devongovett: Yes, that's the right idea. The dir in the app symlinks to a global directory which contains symlinks to individual files/directories in the root of the dependency.

My goal for this feature is to have something that works as closely to `yarn link` as possible but avoids the problem of having multiple copies of dependencies. This is solved by not including `node_modules` in the list of symlinks, and installing the dependencies locally in the project you're linking into.

It's a benefit that it happens to be closer to what you'd actually publish, but not the main goal for me.

bestander: I am not a big fan of multiple symlinks that work around node_modules and .yarnignore.
That would be quite hard to debug in non trivial cases IMHO.

And yeah, it goes the other direction with the `--pack` idea which should be about simulating a publish.

devongovett: Unless I was reading the original knit proposal wrong, I believe this is exactly what was specified.

> A symlink is created for each file or directory that yarn publish would normally have packed up.

See also @ide's comment above:

> If your package has src and build directories and your package.json specifies "files": ["build"], then only the build directory should get symlinked (and shallowly, not the files within it).

devongovett: Following this logic, if you specified `"files": ["build/*.js"]` or something, rather than `["build"]`, we would create a symlink for each `.js` file under `build` instead of the directory itself. This would match the behavior of `yarn pack`. Is there scenario where it would not match?

ide: It sounds like there are two different but related objectives here: better linking that is much closer to doing an actual publish (things like `npm link` that make development different from production lead to a bunch of subtle bugs discovered slowly) vs. fully simulating publishing as if we had a local repo.

For the first (linking), some commands that are intuitive to me are:
```
lib$ yarn link --pack
app$ yarn add lib --link
```

and for the second (local publishing):

```
lib$ yarn publish --local
app$ yarn add lib --local
# or
lib$ yarn pack
app$ yarn add lib.tar.gz
```

bestander: In that case we could split it into two specs then.
I understand the pain that `yarn link` brings when it brings all dev files + node_modules that you don't want.

@devongovett, if this spec is all about symlinking `dep` how about simplifying it a bit?
What if you create the symlinks in `app/node_modules/dep` directly instead of ~/.yarn-links?
My concern is that ~/.yarn-links will grow into having lots of dead links and garbage after some time and it does not play any significant role.

devongovett: I think both of these objectives can be solved by the current proposal, unless you know of a scenario where `yarn link --pack` wouldn't match `yarn pack` other than using symlinks instead of normal files/dirs.

The reason `~/.yarn-links` exists is so you don't have to re-install every time a file is added. Instead, you republish inside the dep you are changing. I think this makes more sense intuitively, and saves time re-installing in every app that uses the dep each time a change to dep is made.

Also, `~/.yarn-links` already exists (it's actually `~/.config/yarn/link/` on my system at least) and it's currently used by the normal `yarn link` command.

Finally, dead links to deleted files would be cleaned up by a re-run of `yarn link --pack`.

bestander: My interpretation of `yarn knit` goal is to have end to end simulation of consuming a first party package as if it was published to a registry, but without the network part and maybe without wasting CPU on gzipping.

I still think Yarn needs this feature because it enables plenty of use cases without having limits by OS, file system and package folder structure.

I recognize that some people don't want to pay the fs.copy performance tax and would compromise on consistency by using `yarn link` or `link:/` as long as there is a way to blacklist `node_modules` and other dev folders.
That is why I think the proposed solution should be separate from the `knit` proposal.

As for "compromise on consistency" I am not convinced that transforming
```
~/work/project_a/node_modules/dep -> (symlink) ../../dep
```
into
```
~/work/project_a/node_modules/dep -> <buffer folder> -> (symlink) ../../dep/file1
~/work/project_a/node_modules/dep -> <buffer folder> -> (symlink) ../../dep/file2
~/work/project_a/node_modules/dep -> <buffer folder> -> (symlink) ../../dep/file3
~/work/project_a/node_modules/dep -> <buffer folder> -> (symlink) ../../dep/file4
~/work/project_a/node_modules/dep -> <buffer folder> -> (symlink) ../../dep/folder1
~/work/project_a/node_modules/dep -> <buffer folder> -> (symlink) ../../dep/folder2
~/work/project_a/node_modules/dep -> <buffer folder> -> (symlink) ../../dep/folder3/file1
...
```
would be easy to implement and maintain.

It feels like it will leave clutter and stale references when used heavily.
That is why I want it to be localized where it is easy to clear and monitor, node_modules of the project seems reasonable.

What would be the reason to make it global?

devongovett: The reason it make it global is so when you re-"publish" (e.g. after adding a new file) you don't also have to re-"install" in every project that uses that dependency. This is because those projects point to the global directory, not to the individual files, so updating the global cascades the updates everywhere.

As for cleaning up references, we have a couple options:

1. We could clean them up when you run `yarn link --pack` again - we'd blow away everything inside the global dir for the package and create new symlinks. This way anything that didn't exist anymore would go away.

2. We could do (1) but across all packages in the global dir as a cleanup step: find any broken links and delete them.

bestander: Makes sense if you to have multiple projects refer the same package.
Is it a common use case for you?
I though that `yarn link` is something that you do temporarily while testing a dependency and don't leave a link for long.

devongovett: Yeah at work we tend to build our apps as a composition of a bunch of libraries, and we have several different apps which all use the some of the same libraries. Setting up and maintaining the links between all of our repos has been a huge pain for us, which is why I'm invested in this feature. Our devs work on different parts of our apps which live across repos all the time, so we want to keep things linked generally. I've written some automation to setup the repos locally for everyone and link everything together using the POC of this yarn feature, and it seems to work pretty well. Ideally we'd use workspaces, but since we have many libraries that are used across different applications, it would be hard to go the monorepo route for us.

bestander: Thanks for clarifying and investing in this feature, @devongovett.
I agree in general that Yarn is in the position to solve your problem but before merging your RFC I want us to sort out a few things that bother me.

1. I want to keep `knit` proposal separate from this one.
I think `knit` is about running `prepublish` scripts and making files available to projects that use a `dep` .
This proposal is about improving `link` command with a trick around symlink so that node_modules and dev files don't leak from `dep` into a project using it.

2. Semantics.
After thinking more, `--pack` seems misleading.
I would say `yarn link --exclude-ignored-files` or something like that, I don't insist on the actual flag name but it should not imply that some "packing" or prepublish will be happening.

3. I am still not convinced that symlinks should be created in `~/yarn-links` or any other global folder.
Symlinks are cheap to create and any issues with the implementation would be contained inside a project that uses this feature.

4. Another thing is this issue https://github.com/yarnpkg/website/issues/573.
We have quite a few way of working with npm packages locally and I am concerned that we are getting too many very similar solutions that approach the problem from different angles.
I suggest hiding this RFC implementation behind a flag, similar to "workspaces-experimental true", and running with it for a while as an opt-in feature. 
So that if we find a better solution we could remove or change it without concerns of breaking the world.

ide: FWIW the linking behavior is what we (Expo) want, whatever the name of the feature ends up being. These are our needs:

(a) we have a repo with multiple apps and multiple libraries
(b) we want the apps to be able to use the latest code in the libraries
(c) we want the libraries to be installed/linked as if we had published them to npm and then installed them in the apps. `npm link`/`yarn link` today has very different behavior than installing published code and we can't trust those commands.
(d) we want to be able to write code in the libraries and easily see how it affects an app. symlinks seem like a good way to do this.

---

Separately, we also do have a need for packing up a library (to simulate publishing) and installing in an app -- this happens in our continuous integration systems. This behavior is even closer to publishing+installing a library, which is why we use it in CI. We run something like: `cd library && yarn pack && cd app && yarn add library.tgz`.

However, in CI we don't need a tight feedback loop between editing library code and seeing the effects of those edits in an app. During development, we want to edit library code and see the app change immediately, which is why symlinks are appealing. We have a solution for CI already (`yarn pack && yarn add`) but no good one for development today.

bestander:
Looks like this PR got stuck :(
I'll be happy to iterate more on the details.

devongovett:
@bestander sorry about that. I got really busy, and a bit confused about what the next steps are for this proposal. However, we really are in desperate need of this feature, so I am happy to work on it if we can come to a well-defined agreement about what we need. 

Here is my proposal to move things forward:

1. Split this proposal out into a separate document, leaving `knit` as it is. We can update the current doc to be much more clear about how the feature should work on some of the finer points discussed in the comments.
2. `yarn link --pack` (or some other name) should run prepublish scripts, and link all files that would be published (following `.npmignore`, excluding `node_modules`, etc.) into a global directory, just as `yarn link` works currently, but linking individual files instead of the project directory.
3. `yarn install --link` installs all dependencies, and for packages that are in the global link directory, those packages are linked.

The global directory is needed so that `yarn install --link` knows which projects are available to link. It also enables changes to be made within a project and automatically propagate those changes to all places that use it.

Please let me know if there are other things that we should cover.

bestander:
Hey, @devongovett, no worries, I can totally understand.
Same here, I got busy with other projects and could not be very responsive recently.

I wish Yarn would have a plugin API so that people could code their needs to get unblocked.

On your proposal.

1. +1 let's make the minimal possible change you need to get unblocked.
It is way easier to get a smaller feature in.
And +1 on updating the current doc with a few comments guiding the future contributors.

2. Ok, we can discuss the flag name and the folder layout in the new RFC.

3. Do we need this flag?
It seems more consistent if `yarn link pkg-a` links a package that was declared with `yarn link` command before.

accepted/0000-link-options.md
```diff
@@ -0,0 +1,93 @@
...
+1. Ran `npm publish` in `dep` (assume that it can clobber an existing version, and that you're publishing to a local registry on just your computer).
+2. Ran `yarn add dep` in `app`.
+
+## Why this behavior is great
+
+This solves several problems that "yarn link" has today:
+
+#### Isolating `node_modules` correctly
+
+You can install `dep` in two different apps without sharing the `node_modules` of `dep`. This is a problem with Electron apps, whose V8 version is different than Node's and uses a different ABI. If you have `node-app` and `electron-app` that both depend on `dep`, the native dependencies of `dep` need to be recompiled separately for each app; `node-app/n_m/dep/n_m` must not be the same as `electron-app/n_m/dep/n_m`.
```
sokra: I don't think the proposal will solve this issue. Node.js calls `realpath` on every "module". So even when files are linked instead of the whole folder, any lookup origining from the linked module will resolve in the `realpath` -> in the `node_modules` folder of the linked dependency and not in the app's node_modules.

devongovett: Not when run with `--preserve-symlinks`: https://nodejs.org/api/cli.html#cli_preserve_symlinks. See drawbacks section. I think webpack already supports such an option as well, and I tried to patch browserify. https://github.com/browserify/browserify/pull/1742

sokra: Ah I see, that makes sense.

arcanis:
This rfc is pretty old, and predate our workspace feature which is a much safer alternative to `yarn link`. I don't think we'll implement `yarn knit`.

SamB:
@arcanis: workspaces may be safer, but they seem to be fairly explicitly intended for *planned* usage, not ad-hoc "oh man, there's an issue in one of our dependencies, let me go make a PR for that, testing it in our thing so I can easily iterate on fixing our actual problem" use cases.

So, features that would _actually_ help with those ad-hoc use cases are still desirable. Since `yarn link` exists and can be used for this, it is the baseline against which other approaches would be measured (with presumably a glance or two at `npm link`), and the advantages weighed against (a) the maintenance burden, (b) the expected cognitive load on users trying to decide what to do, and (c) is it likely that someone will soon discover a much more elegant approach?

**Note**: I have no position on whether any given `yarn knit` proposal would in fact improve things.

[Hmm. I'm really not good at writing short comments, am I?]

PR #74, opened 2017-07-15
Proofread workspaces RFC

valscion:
I found some typos etc. in workspaces RFC doc, so I decided I'd fix them while I was reading the doc.

Looking forward to seeing how workspaces evolve ☺️


bestander:
Thanks a lot!

PR #75, opened 2017-07-23
Upgrading `yarn remove` command

Joklogh:
Command `yarn remove` support deletion of multiple packages

### DIFF
--- a/accepted/0000-yarn-remove-multiple-packages.md
+++ b/accepted/0000-yarn-remove-multiple-packages.md
@@ -1,0 +1,33 @@
+- Start Date: 2017-07-23
+- RFC PR: (leave this empty)
+- Yarn Issue: (leave this empty)
+
+# Summary
+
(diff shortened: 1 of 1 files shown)


cpojer:
Unless I'm missing something, this is already supported. Try `yarn remove jest mkdirp` on the Yarn repository.

Joklogh:
Yeah, you are right, but this is missing in documentation also this feature
is not in add documentation as well. It would be nice if it has better
syntax. For example : yarn add package1,package2 or something like this.
Sorry for adding to rfcs something which is already in yarn.
Also I got off-topic question : how i can write to console? I have tried
console.log() and commander.log() but both do not work.
Thanks kindly for this email.

2017-07-23 12:28 GMT+02:00 Christoph Nakazawa <notifications@github.com>:

cpojer:
Feel free to add this to the documentation by sending a PR to https://github.com/yarnpkg/website. I think the space separated list of packages makes sense both for add and remove and it's unlikely we'll change it to a comma separated list. Thank you for spending time on sending an RFC, however.

PR #76, opened 2017-07-25
Add RFC for install scripts whitelist

voxpelli:
### Background

After discussions turned towards some security related topics in one of the front end chats I'm frequenting and we touched the fact it's [not unheard of](https://arstechnica.com/apple/2016/02/warning-bug-in-adobe-creative-cloud-deletes-mac-user-data-without-warning/) for scripts to accidentally delete peoples hard drives, or to even do something intentionally malicious, we moved the topic of npm scripts and the fact that one either allow them all to run any script they like on install, and exposes oneself to such risks, or one allows none to run anything, and then get a hard time with eg. modules that need to compile some native code.

Pretty much everyone in the discussions felt that it must be possible to do something better so I decided to wrap up some of the thoughts into a suggestion that could be proposed to one of the tools and since I use Yarn, I'm posting it here.

### Suggestion

In summary the suggestion is to introduce the concept of a whitelist to allow a module to run install scripts so that one may vet them before they get to run anything automatically. That way modules that need to run stuff gets to run it after a vetting, the rest doesn't.

Pinging some of the people involved in the front end chat discussions: @aholstenson, @frozzare, @Kolombiken, @murm

### DIFF
--- a/accepted/0000-whitelist-install-scripts.md
+++ b/accepted/0000-whitelist-install-scripts.md
@@ -1,0 +1,76 @@
+- Start Date: 2017-07-24
+- RFC PR: (leave this empty)
+- Yarn Issue: (leave this empty)
+
+# Summary
+
+Ability to whitelist install scripts on a module by module basis.
+
+# Motivation
+
+Install scripts can be problematic from a security perspective, as eg. noted in https://www.kb.cert.org/vuls/id/319816 and [acknowledged by npm](http://blog.npmjs.org/post/141702881055/package-install-scripts-vulnerability). It's also not unheard of that install scripts eg. messes up a recursive delete and deletes unintended data, like eg. [Adobe accidentally did](https://arstechnica.com/apple/2016/02/warning-bug-in-adobe-creative-cloud-deletes-mac-user-data-without-warning/).
+
+Blocking install scripts all together with `ignore-script` though, as is current the only option, also blocks all legitimate use cases, such as modules that needs to compile some code.
+
+From a security perspective it would become more manageable if one could whitelist specific modules and have only the install scripts of those modules be run.
+
+By whitelisting one can vet every module before running its code and decide on a module by module basis whether to allow it or not. Vetting and whitelisting individual modules allows for the full functionality of compiled modules while keeping the security exposure of a project to a minimal.
+
+To enable tools and interfaces to be tried out and patterns to emerge, such that may later be brought into yarn, the proposed design is kept to the minimal necessary core where support is needed within yarn itself.
+
+# Detailed design
+
+For this to work, some things are needed:
+
+1. A way to whitelist the install scripts of a module
+2. A way to enforce such a whitelist
+
+## 1. Whitelist the install scripts of a module
+
+A new flag, `--allow-script`, on `yarn add` and `yarn upgrade` would add an `allowScript` key to that modules entry in `yarn.lock` that tells that the script has its install scripts whitelisted.
+
+## 2. Enforce the whitelist
+
+A new option that's mimicks `ignore-scripts`, but that allows whitelisted scripts (`only-allowed-scripts`) would be a good first step. It would apply both as a global configuration and as a flag on `yarn install`.
+
+Later on it would be good to make it configurable on a project level as well – through `package.json`, `yarn.lock` or somewhere else.
+
+# How We Teach This
+
+As this proposal stands now this functionality would be a fully opt in and not something that a first time user would have to know about.
+
+If it in the future becomes possible to configure a project to have the whitelist imposed on a project level, then it needs to be communicated to new contributors of such a project that such a whitelist is enforced for them so that they when they add or update new dependencies are aware that the whitelist may need to be updated as well for the new dependencies to fully functional.
+
+As this proposal stands now only the options would have to be communicated – how one adds something to the whitelist and how one choses to enforce that whitelist. Not that different from how the functionality of `ignore-scripts` is communicated today.
+
+# Drawbacks
+
+Having a whitelist complicates the list of dependencies and makes the `yarn.lock` more complex.
+
+It also makes the installation process more complex as it becoming harder to know if the install scripts of a module has been run or not as its no longer an all or nothing as is the case of `ignore-scripts`. As it can be somewhat hard today to know which packages that use install scripts and which don't it may to hard to know why a package no longer works as expected. This could be mitigated by clearly communicating whenever an install scripts has been ignored due to not being whitelisted.
+
+As it would become a yarn-only feature it would also leave any users of npm (and of older yarn versions) possibly more vulnerable as less care would be taken to vet the install scripts of non-whitelisted modules. If these however are already using `ignore-scripts`, then things would remain as secure for them.
+
+# Alternatives
+
+## Private registry
+
+In some previous discussions on similar matters, https://github.com/node-forward/discussions/issues/29#issuecomment-70438448, it was noted that private registries can be used to achieve a similar vetting of modules in there and one could probably also modify packages in such a private registry to remove any install scripts one would prefer not to run. Such private registries are though not easily accessible to people + adds extra complexity to a setup that makes it harder for people to contribute and to eg. whitelist new modules through pull requests and such.
+
+# Unresolved questions
+
+## Where to store whitelist
+
+Should it be in the `yarn.lock` or somewhere else, like in the `package.json`?
+
+## Project specific configuration
+
+Should enforcement of whitelist solely happen through flag and global configuration or should a project be able to enforce it at a project level as well, embedding the enforcement in eg. the `yarn.lock` or `package.json` so that everyone else who installs the project with yarn automatically gets the whitelist enforced?
+
+## How to handle the need of whitelisting a sub-dependency
+
+A module like [bunyan](https://github.com/trentm/node-bunyan) which has a module that needs compilation ([dtrace-provider](https://github.com/chrisa/node-dtrace-provider)) as a dependency would have to have that dependency whitelisted rather than to be whitelisted itself. As one only do `yarn add` on the top level dependency there would be no way of whitelisting such a sub-dependency through any existing command so either a new command for whitelisting a specific dependency and/or an extension of `upgrade-interactive` as well a new `add-interactive` would be needed to allow whitelisting of such sub-dependencies.,
+
+An extended `upgrade-interactive` and a new `add-interactive` would make it easy to whitelist sub-dependencies and to ensure that updates to whitelisted modules are vetted before they get whitelisted. An `add-interactive` could point out any non-whitelisted sub-dependency that wants to run an install script and ask whether that should be allowed and whitelisted or whether that should be ignored. If ignored, then that should probably be explicitly saved as well so one don't get bombarded with the same question again and again. Like saving an explicit `allowScript: false` to `yarn.lock` for that module.
+
+By keeping the design of this out of the initial proposal the hope is that experimentation with tools and interfaces for handling these cases can emerge outside of yarn and that yarn later, if the need remains, can bring the best of such experiments into core.
--- a/accepted/0000-whitelist-install-scripts.md
+++ b/accepted/0000-whitelist-install-scripts.md
@@ -4,7 +4,7 @@
 
 # Summary
 
-Ability to whitelist install scripts on a module by module basis.
+Ability to opt-in install scripts for just select modules
 
 # Motivation
 
@@ -12,9 +12,9 @@ Install scripts can be problematic from a security perspective, as eg. noted in 
 
 Blocking install scripts all together with `ignore-script` though, as is current the only option, also blocks all legitimate use cases, such as modules that needs to compile some code.
 
-From a security perspective it would become more manageable if one could whitelist specific modules and have only the install scripts of those modules be run.
+From a security perspective it would become more manageable if one could opt-in specific modules and have only the install scripts of those modules be run.
 
-By whitelisting one can vet every module before running its code and decide on a module by module basis whether to allow it or not. Vetting and whitelisting individual modules allows for the full functionality of compiled modules while keeping the security exposure of a project to a minimal.
+By opting-in one can vet every module before running its code and decide on a module by module basis whether to allow it or not. Vetting and opting-in individual modules allows for the full functionality of compiled modules while keeping the security exposure of a project to a minimal.
 
 To enable tools and interfaces to be tried out and patterns to emerge, such that may later be brought into yarn, the proposed design is kept to the minimal necessary core where support is needed within yarn itself.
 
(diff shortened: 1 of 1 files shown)


bestander:
I like the idea.
Not sure which place is better for the whitelisted scripts - yarn.lock or package.json.

We treat package.json as a blueprint for the project and yarn.lock is more like a table of pattern to URL and we don't have much more features inside it.

voxpelli:
It seems like there's actually right now a case of such a malicious package being used in the wild to steal peoples environment variables: https://twitter.com/o_cee/status/892306836199800836

Great to hear you like the idea @bestander 👍 

I guess one question that impacts where the whitelist goes is what data it should actually contain. Should it whitelist all versions for a package or just a specific version range?

Whitelisting version ranges would be the most powerful. One way to do so could be to just mimic how dependencies are added and do something like this in `package.json`:

```json
"allowScripts": {
  "package-name": "1.0.0 - 1.2.0",
  "another-package": "*"
}
```

The existence of an `allowScripts` key in `package.json` could then also be what signals if a whitelist should be enforced or not, so if one wants to exclude all scripts from running, then one would simply add `"allowScripts": {}`.

Such a property could also be possible for `npm` and other tools to support.

BYK:
We can also move this into `.yarnrc` so we can have global, user-level and project-level granularity?

Daniel15:
Would the whitelist be for a particular version of a package, or for any version of the package? If it's not version-specific, it wouldn't protect against malicious code being added to an existing package, right?

ljharb:
Can we use a slightly less problematic term than "whitelist"? Perhaps "inclusion list"?

bestander:
> Can we use a slightly less problematic term than "whitelist"? Perhaps "inclusion list"?

Why whitelist is problematic?

ljharb:
@bestander "whitelist" is good, "blacklist" is bad, which relies on the connotations of one color being good and one bad, which has racial implications. (i really really hope nobody gets into a debate here on this, if anyone doesn't agree with the premise, a simple "no thanks" would be better than devolving into a debate on whether people's feelings matter)

voxpelli:
@ljharb Right, from my point of view I'll happily switch the wording to "install script opt-in" then, would convey the concept about as well and avoid anyone feeling offended.

As a property in a file I think "allowScripts" make sense, so this would only change the RFC wording I think.

@BYK Interesting idea, `.yarnrc` could work very well.

One benefit of that over `package.json` is that it would convey that dependencies themselves can't opt-in any install scripts.

Another is that it would also make it easy to share an opt-in list between projects, so that a team or company could enforce the same list across many projects.

vmasto:
Given that naming a malicious package similarly to a popular one is a very common attack vector, and as an extra safety blanket, what if `yarn` could keep a global centralized whitelist based on, say, the top 500 most popular packages and before installing it warned for typos based on Levenshtein distance (or any other suitable algorithm).

I also just found this discussion on `npx` https://github.com/zkat/npx/issues/66 which is relevant and discussed the same concept.

tdsmithATabc:
I was going to suggest that it would be nice if you could opt-in to the script "content" itself (for a specific package) so that you didn't have to reapprove it when there had been no changes. The opt-in list would be hashes of the scripts themselves.

But then I convinced myself that this opens up too many weird security edge cases, like a package changing structure in such a way that a previously harmless script became malicious.

Just thought I'd share that one-person conversation. 😛

peabnuts123:
Allowing only packages at specific semvers is, theoretically, ideal. However, it is totally impractical in reality. Direct dependencies are already specified at specific version ranges, and the version of any subdependencies are specified by their respective `package.json` manifests. This means that while you would be able to specify only the packages you would like to include, the packages you would be receiving is ultimately up to the developers of the packages you depend on, the packages they depend on, the packages those packages depend on, etc. Essentially creating an exponential amount of overhead upkeep for a project maintainer, the more packages you depend on. This quickly balloons out of control, my work project containing roughly 1000 unique subdependencies directly in `node_modules`, so, not accounting for version conflicts nested within other dependencies, something you would also have to account for in such an inclusion manifest. This is all also ignoring the fact that you have no reason to trust a package at a certain version anyway, and for a real testament to security, each package would need to be audited individually, at each version. 

This is a serious problem. One that really needs to be solved. However, it is not easily solved, and I don't think an inclusion list is practical, meaning it will do little for the security of the general populace of developers using NPM / Yarn, and only really serve the diehard security-conscious types, who probably have alternatives such as "don't use a package manager" to remedy their concerns already.

voxpelli:
@peabnuts123 I think an opt-in list can be manageable, at least in some cases. I too have a lot of unique modules in my work projects, but very very few of them needs to run any install scripts.

For me personally I would probably run with something like this (in maybe `.yarnrc`):

```json
"allowScripts": {
  "node-sass": "*",
  "gc-stats": "<=1.0.2",
  "dtrace-provider": "<=0.8.3"
}
```

That would narrow down the modules with vulnerable install scripts from a couple of hundreds to just a few, which would make me feel much more safe and which would be much more practical for me than todays only option, `ignore-scripts`.

Of course things can be improved even further. This opt-in list can easily live side-by-side with other security measures and doesn't stop any work from happening on any such ones.

To me this would just be a small simple step towards making install script security a bit more practical and a step that can be iterated on in the future – with eg. addition of better tools, both within yarn itself and within the wider ecosystem.

rally25rs:
Personally, I'd like to be able to just set `ignore-scripts` at the global `.yarnrc` level, but still have my projects able to run scripts for the few packages that are known to need them (PhantomJS for example). This would protect against fat-fingering or typo installs in new projects while still letting existing projects work.

Wherever this is stored, we should be sure to only respect this setting for the current project where Yarn is executing. In other words, an installed dependency should not be allowed to specify an `allowScripts` list in it's `package.json` (so that malicious packages can't add to the whitelist). That is probably obvious, but should be explicitly unit tested so as not to allow a regression where the whitelist can be modified by dependencies.

I don't really like the idea of the community having to maintain a list of "bad" packages, because that is always reactionary. Someone has to realize it, add it to the list, and deploy the list. What if a "bad" package is already in my cache and I'm running --prefer-offline? how do I get the updated list? Just seems like a maintenance nightmare and should really be handled by NPM revoking known malicious packages from the registry.

benyitzhaki:
i agree with @rally25rs regarding NPM's responsibility here. i think they should be able to revoke malicious packages or at least the other way around, flag "trusted packages" and make npm recognize only trusted sources in their repository by default. this could be done as a POC in yarn as from my understanding, they use a proxy for the official repo

voxpelli:
Lets keep to the topic of this RFC though @benyitzhaki, because as stated earlier – nothing in this RFC is incompatible with any such measures from npm:s side. Rather it would be great to have multiple layers of security.

@rally25rs Yeah, version ranges in the opt-in list should rather be for targeting a range of earlier versions than for a range of upcoming ones (even though even such a range can, if wide, leave some room for malicious updates to earlier versions – like patch updates of old minor releases)

I like the idea of using `.yarnrc` as it makes it clear that only the `.yarnrc` within the project folder and above will be accounted for – which means any within a dependency will be ignore, which is important as you point out @rally25rs.

Using `.yarnrc` would also give certain inheritance rules out of the box I guess, which would mean that the `allowScripts` / `allow-scripts` option (latter follows `.yarnrc` naming conventions better) defined in the `.yarnrc` closest to the project folder would be the one that's used.

One thing regarding such inheritance though that would need to be defined is how `ignore-scripts` and `allow-scripts` would interact.

I would say that whenever `allow-scripts` is defined `ignore-scripts` should be ignored. The only possible issue with that from a backwards compatibility perspective would be if a local project sets `ignore-scripts: true` to deactivate all scripts locally and there's a `allow-scripts` set globally – then that local deactivation will be ignored until they override the `allow-scripts` instead.

bestander:
@xMrWhite, good idea.
From implementation POV could be harder because Yarn runs scripts in parallel and prompting in scripts is not supported

xMrWhite:
It should be able to determine all the packages that have script that could be run during the resolution phase. The scripts should continue to be run in parallel. I guess it depends on if the full package.json is queried during resolution.

voxpelli:
@xMrWhite I was thinking about something similar as well, but decided to keep this RFC focused on the minimal possible implementation so that it would be easier to later on try out different such interfaces and enable work on multiple such approaches in parallel, without any single one holding up the core mechanism itself.

(For some such a mechanism that you suggest would probably be ideal, for others there can perhaps be a team-wide opt-in list instead, all depending on the type of project and the composition of development team. The sooner we can start experiment with different such interfaces, the better)

rally25rs:
I like the idea of prompting per script, but it would have to be an opt-in to stop and prompt. Keep in mind a lot of `yarn install`s are done by automated systems, like CI servers. Would probably have to be something like an `add-interactive` or `install-interactive` command, like we do with `upgrade` and `upgrade-interactive`.

My thought for an initial implementation would be to just add a `name:version` hash (same format at `dependencies` in package.json) to .yarnrc and those scripts are executed even if `--ignore-scripts` is passed.

xMrWhite:
It should go into the yarn lock for the project, that already contains version information

bestander:
Good point raised, @xMrWhite.
Yarn guarantees consistent node_modules for the same yarn.lock + package.json + execution options (--production etc).

The role of yarn.lock is to store resolutions for all package patterns used in the project.
I don't think we should overload it with a new role that is not related to dependency resolution.

.yarnrc sounds like the right place to store execution option, package whitelist is one of them.

rally25rs:
Maybe we should discuss the workflow.

Would we expect a user to already have `ignore-scripts` set in their `.yarnrc` then do something like `yarn add foo --allow-scripts`? If that is the case then maybe `yarn.lock` would make sense, but I bet a lot of people just manually add entries to `package.json` and run `yarn install`. In that case, then couldn't manually edit `yarn.lock` to enable scripts for that package because an entry wouldn't exist yet.

If someone already decided to disable scripts, then they probably already have a `.yarnrc` file with `ignore-scripts` in it so it could be manually edited to add vetted packages to an array.

peabnuts123:
If anybody wants to audit what dependencies have `preinstall`, `install` or `postinstall` hooks you can run this little script in a terminal:

```
find node_modules/ -iname "package.json" -exec grep '"preinstall"\|"install"\|"postinstall"' {} \; -exec echo 'From: {}' \; -exec echo \;
```

Output:
```
    "install": "node install",
From: node_modules//fsevents/package.json

    "install": "node scripts/install.js",
    "postinstall": "node scripts/build.js",
From: node_modules//node-sass/package.json

    "postinstall": "node postinstall",
From: node_modules//spawn-sync/package.json
```

in my IDE (vscode) I can just click these file references and see exactly they are invoking, etc.

voxpelli:
This issue seems to have come up again: https://github.com/eslint/eslint-scope/issues/39 A module stealing credentials (I think on install)

I dropped the ball on this one due to life getting in the way pretty majorly, sorry 😐 I’ll try to pick this up again soon and rewrite the proposal to include the given feedback.

RumataEstor:
Would an approach to allow specific features be more convenient?

For example, when a script is executed, a custom "require", "fs", "http" and whatever else is provided. Whenever they are executed, they check if the requested operation has been already allowed in the yarn.lock previously, or if not would ask the user interactively if it's safe to execute.

The allowed operations might be like "fs read relative_path", "http get url_wildcard" and so on.

dominykas:
Implemented this as a separate package: https://www.npmjs.com/package/allow-scripts

tl;dr:

- add [`allowedScripts: { pkg: versionRange }`](https://github.com/yarnpkg/rfcs/pull/76#issuecomment-319614910) in your `package.json`
- `npm install --ignore-scripts` or `yarn install --ignore-scripts`
- `npx allowed-scripts`

voxpelli:
Awesome @dominykas! Thanks for picking this up

I see that nvm is showing interest in this: 
https://github.com/yarnpkg/rfcs/pull/76

I’ll happily try to update this PR with whatever the conclusion is there, or if there is none, whatever we would like to push here instead 👍

arcanis:
Fyi the v2 includes a way to disable postinstall scripts, via `dependenciesMeta`:

```json
{
  "dependenciesMeta": {
    "webpack-cli@1.2.3": {
      "build": false
    },
    "webpack-cli": {
      "build": false
    }
  }
}
```

dominykas:
@arcanis does that mean it has them enabled by default?

arcanis:
Yes, even though I was considering disabling them by default. This will likely be a rfc since it has far reaching consequences. 

Whatever we end up doing, the end goal is to give you the ability to choose whether they should execute or not both on a global basis and a per-package basis.

arcanis:
> Will the work in v2 be synced with npm in any way?

Not particularly. Working with npm proved emotionally exhausting for me during this past year, and I've decided to stop being the only one of us [making efforts to be the good neighboor](https://twitter.com/arcanis/status/1040646640967794690), [waiting months for them to share their wisdom](https://github.com/yarnpkg/rfcs/pull/105), or [seeing our features rebranded without so much as a notice](https://github.com/npm/rfcs/pull/27).

Even now I'd be happy to include us and discuss with them as peers, but it takes two to tango and I've lost hope seeing it happening under the present circumstances. npm be npm.

> Should this RFC be updated in any way to take the work you mention into consideration?

That sounds like a good idea - would you like me to update it? Or you can pop into [Discord](https://discord.gg/vkrvmaW) to discuss it, both are fine by me 🙂

chocolateboy:
@voxpelli Please change the wording back. The [latest edit](https://github.com/yarnpkg/rfcs/pull/76/commits/c8c91a88e308823e322ae45a35bb2560c1b1651d) is absurd.

> "whitelist" is good, "blacklist" is bad, which relies on the connotations of
> one color being good and one bad, which has racial implications.

There are no racial connotations to "blacklist" any more than there are to "black mark" or "[denigrate](https://english.stackexchange.com/a/209032)". As a community, developers tend to be reasonably vigilant about calling out woo and bad science. It's a shame we don't exhibit the same vigilance when confronted with woke woo and bad etymology (a subset of bad linguistics).

AFAICT, the commenter suggesting the change hasn't contributed to this project. They are not its gatekeeper, and complying with this passive-aggressive proposal is not a prerequisite to getting this RFC accepted.

> if anyone doesn't agree with the premise, a simple "no thanks" would be [good]

Agreed.

> [rather] than devolving into a debate on whether people's feelings matter

**Strongly** disagree with the insinuation that anyone who objects to this absurd change believes that feelings don't matter. What gives you (a white person AFAICT) the authority to speak for everyone else, including me (who isn't white), and to proactively seek out and take an offence that no-one intended or inferred prior to your intervention?

If the motivation for this change was to avoid causing offence, it has massively failed. We (black people) don't need white people with a savior complex to protect us from words. By focusing on nonsense like this, all they do is succeed in infantilizing people of color and drawing attention away from *actual* racial discrimination and racially-aggravated violence.

voxpelli:
@chocolateboy I do agree with you, I neither find any racial connotations to "blacklist" but I did want to avoid derailing this issue into a discussion on that matter.

I do think the new wording isn't that much worse (or any better) so I do think it can remain and we can move this issue beyond such distractions, if that would that be okay with you? Reverting it would just risk derailing this even further.

I do think this discussion, and the points that you raise, could be a very good one to have, but it would preferably be in another forum, like a blog, where such discussions can be held independently on any RFC in itself. Can you agree @chocolateboy? And you @ljharb?

ljharb:
@chocolateboy it’s great for you that you find no such connotations; every other human doesn’t share that opinion. The change is not absurd, whether you agree with the motivations or not, and yes, objecting to it does say things about you whether you want it to or not.

@voxpelli i agree that this isn’t the forum to be debating it, but it’s also something that’s been debated in many other forums, and I’m not the origin or the sole impetus for this particular kind of change. I’m glad you agree that leaving the edit as-is is the best choice for the scope of this rfc.

voxpelli:
> it’s also something that’s been debated in many other forums

Assuming that everyone else have taken part of those debates and thus that there is no need to reference them is rather paradoxically not a very inclusive way of dealing with such a question in my opinion.

It doesn't benefit the cause but rather just risks derailing the debate into something that everyone would prefer it wouldn't become.

So maybe next time reference some of those debates or prior art then @ljharb? So people can know where to take those discussions 🙏

ljharb:
I’m happy to google for you; https://www.ncbi.nlm.nih.gov/pmc/articles/PMC6148600/ seems relevant, and https://github.com/facebook/react/issues/13604 is a similar change in react.

arcanis:
I'd like to ask everyone to assume good intent when addressing someone else. If you disagree with something someone else says, please consider that they believe that what they're doing is the right thing. In this context, using strong words like "absurd" is best avoided.

**Anyway,** I believe the use case for the feature described in the RFC has been solved under the form of the new field [`dependenciesMeta.built`](https://yarnpkg.github.io/berry/configuration/manifest#dependenciesMeta.built), currently exclusive to the v2. Together with `enableScripts` it can be used to form either of the two common modes (exclusion list / inclusion list), and supports targeting packages by name and versions alike. What do you think, @voxpelli?

voxpelli:
Very neat @arcanis 👌 I would very much say that it solves this and it does so very elegantly! Thanks 🎉

eawww:
@voxpelli @ljharb @chocolateboy and whomever else: I personally think this was a missed opportunity to shift the convention from "blacklist" and "whitelist" to "naughtylist" and "nicelist". Racial connotations or no, it's just superior nomenclature.

PR #78, opened 2017-08-02
Scripts whitelist

rally25rs:
added scripts-whitelist.md for comment.


rally25rs:
Deleting this because another RFC already exists.

PR #80, opened 2017-08-23
Add yarn -v

AleshaOleg:
When i type in console `yarn -v` I expect to see installed version of yarn (like with `yarn --version`), as in most of other command applications. Instead, as I understood, if command not determined, yarn runs `yarn install`.


arcanis:
Already fixed in https://github.com/yarnpkg/yarn/commit/21566ddb5c9fd07b94a0b92390df19911b994bd5, will be released in the 1.0.

PR #81, opened 2017-09-12
Create 0000-yarn-global-run

Haroenv:
see also https://github.com/yarnpkg/yarn/issues/4392

[rendered](https://github.com/Haroenv/rfcs/blob/patch-1/accepted/0000-yarn-global-run.md)


bestander:
Is it similar to `npx`?

Haroenv:
It definitely is

accepted/0000-yarn-global-run
```diff
@@ -0,0 +1,43 @@
...
+
+Npx already exists and fills this position for now. 
+
+Adding a flag like `--exact` to `yarn create` so that the `create-` prefix isn't 
+automatically added
+
+# Unresolved questions
+
+- will the installed package be installed globally
+- is the name good
```
BYK: I think we should add an item about a possible consolidation with `yarn create`

accepted/0000-yarn-global-run
```diff
@@ -0,0 +1,43 @@
...
+# Alternatives
+
+Npx already exists and fills this position for now. 
+
+Adding a flag like `--exact` to `yarn create` so that the `create-` prefix isn't 
+automatically added
+
+# Unresolved questions
+
+- will the installed package be installed globally
```
BYK: `yarn create` installs it globally and doesn't clean it up after the run for future invocations. I expect this to behave the same unless people object.

accepted/0000-yarn-global-run
```diff
@@ -0,0 +1,43 @@
...
+# Drawbacks
+
+It can cause confusion with regard to `yarn create`, since it has a very similar
+goal. 
+
+# Alternatives
+
+Npx already exists and fills this position for now. 
+
+Adding a flag like `--exact` to `yarn create` so that the `create-` prefix isn't
```
BYK: Can you think of more? May be a separate binary like `yr`?

accepted/0000-yarn-global-run
```diff
@@ -0,0 +1,43 @@
...
+`create-`. 
+
+# How We Teach This
+
+Yarn run and Yarn global already exist, so this is not introducing new concepts, 
+especially since `yarn create` already exists.
+
+# Drawbacks
+
+It can cause confusion with regard to `yarn create`, since it has a very similar
```
BYK: May be we can come up with some safety measure for `yarn create` that limits damage/impact for scripts adhering to certain standards and `yarn global run` becomes the "escape hatch" everything else?

One thing that I can think of is not making binaries from `yarn create` installs globally available at all so you __have to__ run `yarn create react-app` even after installing it.

accepted/0000-yarn-global-run
```diff
@@ -0,0 +1,43 @@
...
+
+# Detailed design
+
+The design would be exactly the same as `yarn create`, but without the prefixing of
+`create-`. 
+
+# How We Teach This
+
+Yarn run and Yarn global already exist, so this is not introducing new concepts, 
+especially since `yarn create` already exists.
```
BYK: We may wanna differentiate it a bit more. Also, a blog post would probably be helpful along with docs.

BYK:
I think this is an interesting and good idea. What do you think @yarnpkg/core

thejameskyle:
I'm not sure about the name `run`. When I first read `yarn global run` I expected it to work exactly like the `yarn run` command.

accepted/0000-yarn-global-run
```diff
@@ -0,0 +1,52 @@
...
+
+Some packages that are similar in design to `create-react-app` ([vue-cli](https://yarn.pm/vue-cli),
+[preact-cli](https://yarn.pm/preact-cli), [@angular/cli](https://yarn.pm/@angular/cli)
+...) don't have a name that's prefixed with `create-`, and thus prevent the usage of
+`yarn create`, even though it would be perfectly suitable for it. 
+
+# Detailed design
+
+The design would be exactly the same as `yarn create`, but without the prefixing of
+`create-`.
```
thejameskyle: Could this replace `yarn create`?

Haroenv: I guess that's a option

ljharb:
npx is nice because it's x like "exec".

ljharb:
(to clarify https://github.com/yarnpkg/rfcs/pull/81#issuecomment-329871642, i was suggesting that the name be `yarn exec`)

bestander:
I'll be boring and suggest considering other options like a separate package or promote an alias in bash for that.
Less features, more time we can spend on maintaining others :)

Haroenv:
this is long superseded by yarn dlx

jerrygreen:
I tried `yarn dlx` but that doesn't seem to work

PR #82, opened 2017-09-12
Adding RFC file for pinning yarn version in a project

psbanka:
Hello, given that our CI system requires yarn version `0.17.8` and that we have developers on a dozen different versions of yarn, we thought it would be helpful to have yarn communicate to our developers exactly what version of yarn they should be using. This RFC provides a direction for how this might be done.

### DIFF
--- a/text/0000-pin-yarn-version-in-yarnrc.md
+++ b/text/0000-pin-yarn-version-in-yarnrc.md
@@ -1,0 +1,95 @@
+- Start Date: 2017-09-12
+- RFC PR: 
+- Yarn Issue: 
+
+# Summary
+
+Since different versions of yarn have different guarantees as to how `node_modules` are installed, allow a project to select
+what version of yarn should be used for that project.
+
+# Motivation
+
+Currently when running yarn, the system warns the user whenever the version
+of yarn is not the latest version. However, there may be reasons _not_ to use
+the latest version (e.g. incompatibility with your CI system), or errors that
+are [introduced by a new version](https://github.com/yarnpkg/yarn/issues/4404)
+of yarn that impact your project.
+
+Instead, a project should be able to designate a version of yarn that all
+developers should be using for that project and it should warn them if they
+are using a version of yarn that is not the same as what that project
+expects.
+
+# Detailed design
+
+Add an entry into `.yarnrc` which specifies the version(s) that are acceptable to be used, perhaps using the same format as `package.json` itself, e.g.:
+
+```
+project-version ^1.0.2
+```
+
+If the user is running `yarn install` without the `--prod` flag, and there is
+a version-mismatch, yarn should generate a warning and tell the user to
+download the latest acceptable version. If `yarn install` is running _with_
+the `--prod` flag, and there is a version-mismatch, yarn should error and
+quit (stating that there is a version-mismatch).
+
+# How We Teach This
+
+**A common criticism of yarn is that it [does not guarantee determinism between
+different versions of yarn](https://news.ycombinator.com/item?id=14479693)
+and there is no way to determine in a given development team if any individual engineer is using the correct version of yarn. Now if a larger team wants
+to guarantee that, they can.**
+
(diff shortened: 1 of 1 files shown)


text/0000-pin-yarn-version-in-yarnrc.md
```diff
@@ -0,0 +1,95 @@
...
+of yarn that impact your project.
+
+Instead, a project should be able to designate a version of yarn that all
+developers should be using for that project and it should warn them if they
+are using a version of yarn that is not the same as what that project
+expects.
+
+# Detailed design
+
+Add an entry into `.yarnrc` which specifies the version(s) that are acceptable to be used, perhaps using the same format as `package.json` itself, e.g.:
```
bestander: I wonder if this setting could better be in package.json instead?

bestander:
Seems reasonable, I think it could be useful.

bestander:
What do you think, @arcanis  @BYK?

BYK:
I have been wondering why we are not using the already existing and supported `engines` field in `package.json`?

BYK:
Btw. thank you very much for all the effort you've put into submitting this @psbanka! I'm not trying to be dismissive. I've just been wondering that question and I'm probably missing something.

arcanis:
If you want to warn them about the mismatching Yarn version, you can use the `engine` field as @BYK suggested.

If you want to force them to use a specific version, you can use the `yarn-path` settings in the `yarnrc` file. This is not yet documented (PR wanted!), but you can check [the PR that introduced it](https://github.com/yarnpkg/yarn/pull/4204).

psbanka:
Makes sense. Sounds like the consensus is to use the `engine` field. Would it be helpful if I modified this RFC to specify this?

BYK:
>Makes sense. Sounds like the consensus is to use the engine field. Would it be helpful if I modified this RFC to specify this?

I'd rather have a short guide/article on the website explaining how to achieve this. Would you be interested in preparing something?

psbanka:
@BYK – I'd be happy to write up such a guide/article, but are you suggesting that the `yarn-path` setting is the solution for this? The concerns I have about this approach are:

1. specifying a straight path does not necessarily indicate version (what version is `/usr/local/bin/yarn`, for example?

1. Since there is a wide variety of different ways of installing `yarn` (e.g. brew, npm, curl), all of which may result in an installation to different paths, you cannot specify a path without also requiring devs to also specify a specific means on installation. This seems like an unnecessary complication.

1. Even if we were to be able to surmount the previous two issues, this solution wouldn't provide the same kind of functionality as `engines`, where we would specify a range of different potential yarn versions that are acceptable.

rally25rs:
For reference, this would be helpful to avoid things like this: https://github.com/travis-ci/travis-ci/issues/7566

rally25rs:
My opinion on this is that Yarn should check `engines.yarn` in package.json (as @BYK mentioned above) as this is what NPM does https://docs.npmjs.com/files/package.json#engines

> You can also use the "engines" field to specify which versions of npm are capable of properly installing your program. For example:
>
>{ "engines" : { "npm" : "~1.0.20" } }
>
>Unless the user has set the engine-strict config flag, this field is advisory only will produce warnings when your package is installed as a dependency.

szemek:
I have a question a bit related to this topic. Have you considered storing `yarn` version in `yarn.lock`? In Ruby ecosystem, `bundler` is storing its version in `Gemfile.lock`.
One thing is define acceptable versions, but it might be valuable to have an exact information about used version.

rally25rs:
I actually hate that about bundler. It causes the Gemfile.lock to change between every dev that touches the project, unless the entire team is on the exact same version, which in practice doesn't happen that often (usually the same major, sometimes minor version, but seldom is everyone on the same patch version of bundler).

bestander:
@szemek, Yarn even had this feature for a very short time, it was reverted after people started seeing too many merge conflicts.

szemek:
@bestander thanks for explanation.

PR #83, opened 2017-09-21
RFC for running workspace commands from root

giridharangm:
Redirected from [here](https://github.com/yarnpkg/yarn/issues/4467#issuecomment-330905517)  

It would be great if we have a single command to execute a package script for all the packages under a workspace.

### DIFF
--- a/accepted/0000-workspace-run-commands.md
+++ b/accepted/0000-workspace-run-commands.md
@@ -1,0 +1,57 @@
+- Start Date: 2017-09-21
+- RFC PR:
+- Yarn Issue:
+
+# Summary
+
+Allow Yarn CLI to execute a package script on each packages from the workspace root.  
+
+# Motivation  
+
+"Lerna". Lerna does a great job for handling monorepos. Since Yarn has a built-in 
+workspaces feature, we could use some of the functionalities that being used in 
+lerna (not entirely though!) for handling the monorepo more efficiently.  
+
+Just like installing all dependencies from one place (workspace root), it'd be great 
+to execute all the sub-package scripts from the workspace root using a single command. 
+Not only this would make a less back-and-forth traveling between the packages to 
+execute scripts, but also it'll help a lot for CI/CD configuration. 
+
+
+# Detailed design  
+
+As @BYK mentioned [here](https://github.com/yarnpkg/yarn/issues/4467#issuecomment-330873337), 
+we could create two commands for this feature.  
+
+* `yarn workspaces list`  
+* `yarn workspaces run <command>`
+
+By introducing `workpaces` command, it brings a modular approach from the CLI's perspective.  
+
+## `yarn workspaces list`  
+
+This command lists out the package names inside that workspace.  In the case of multiple workspace configuration, we'll need to pass the workspace name like `yarn workspaces <workspace_name> list`. By defaut, when no workspace name is not given, it'll pick the first workspace name from the root package.json.  
+
+## `yarn workspaces run <command>`  
+
+This command will execute the specified command in all the packages under the workspace. Before we run the scripts, it should make sure that all packages has the script specified in the package.json. Otherwise, it should throw an error specifying which pacakge is missing that script.  
+
+Just like the `list` command, by default, it'll select the first workspace in case of no workspace name is provided.  
+
+The important thing here to note that the ordering of the execution. It must be executed in the same dependency order. This will assure that we could use a single "build" that builds inter-dependant packages without any problem.  
+
+# How We Teach This  
+
+Since this is a new feature and doesn't affect existing functionalities, it shouldn't affect existing/new users. However, a slight documentation should be written under the workspace CLI.  
+
+# Drawbacks
+
+Complexity. Especially while executing the package scripts. We need to keep track of the package dependencies.  And also, there's already a library (lerna) does the same thing and is compatible with yarn without any extra configuration.  
+
+# Alternatives
+
+Don't have an alternative way (I can think of) inside yarn.
+
+# Unresolved questions
+
+How to handle multiple workspaces?
--- a/accepted/0000-workspace-run-commands.md
+++ b/accepted/0000-workspace-run-commands.md
@@ -24,8 +24,6 @@ we could create two commands for this feature.
 
 * `yarn workspaces list`
 * `yarn workspaces run <command>`
-* `yarn workspaces add <workspace_name>`
-* `yarn workspace remove <workspace_name> --opts`
 
 By introducing `workpaces` command, it brings a modular approach from the CLI's perspective.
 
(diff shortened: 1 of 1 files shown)


accepted/0000-workspace-run-commands.md
```diff
@@ -0,0 +1,57 @@
...
+
+# Detailed design  
+
+As @BYK mentioned [here](https://github.com/yarnpkg/yarn/issues/4467#issuecomment-330873337), 
+we could create two commands for this feature.  
+
+* `yarn workspaces list`  
+* `yarn workspaces run <command>`
+
+By introducing `workpaces` command, it brings a modular approach from the CLI's perspective.
```
BYK: Possible additions:

*yarn workspaces add <name>*: This would create the folder if it doesn't exist and make sure it is included in the top-level `workspaces` config.
*yarn workspaces remove <name>*: Remove a specific workspace from the config and optionally delete the folder if `-D` flag is passed.

accepted/0000-workspace-run-commands.md
```diff
@@ -0,0 +1,57 @@
...
+we could create two commands for this feature.  
+
+* `yarn workspaces list`  
+* `yarn workspaces run <command>`
+
+By introducing `workpaces` command, it brings a modular approach from the CLI's perspective.  
+
+## `yarn workspaces list`  
+
+This command lists out the package names inside that workspace.  In the case of multiple workspace configuration, we'll need to pass the workspace name like `yarn workspaces <workspace_name> list`. By defaut, when no workspace name is not given, it'll pick the first workspace name from the root package.json.
```
BYK: Can you clarify what you mean by a `multi-workspace` config?

giridharangm: I'm so sorry.. Got confused with this. 
image I thought we had a multi-workspace setting. Will remove that.

accepted/0000-workspace-run-commands.md
giridharangm: For the `yarn workspaces add`, shouldn't we pass the workspace name too?

kaylieEB: I believe yarn workspaces run <command> is already implemented via https://github.com/yarnpkg/rfcs/blob/master/implemented/0000-workspaces-command.md#yarn-workspace-workspace-name-command-name-?

jharris4:
This feature could be useful to help clean files from individual packages in workspace, such as `node_modules`.

Some related issues from non-workspace yarn are:

https://github.com/yarnpkg/yarn/pull/1284
https://github.com/yarnpkg/yarn/issues/1048

James-E-Adams:
What are other peoples' thoughts on being able to execute a particular package's script in isolation?

I.e If `workspace-a` has a `test` script, being able to run `yarn workspace-a:test`, rather than run it for every workspace.

accepted/0000-workspace-run-commands.md
```diff
@@ -0,0 +1,67 @@
...
+The important thing here to note that the ordering of the execution. It must be executed in the same dependency order. This will assure that we could use a single "build" that builds inter-dependant packages without any problem.  
+
+
+## `yarn workspaces add <workspace_name>`  
+
+This would create the folder if it doesn't exist and make sure it is included in the top-level `workspaces` config.  
+
+## `yarn workspace remove <workspace_name> --opts`  
+
+Remove a specific workspace from the config. And by passing `-D` as an option, it'll delete the folder.
```
bestander: For simplicity I would focus only on list/run commands.

Add/delete are less utility and more like sugar on top of editing package.json and adding/removing folders.
Also I don't like inconsistency workspaces/workspace for those commands
And in Yarn we prefer using explicit flags and not shorts `--delete-folder` would be more consistent with other commands

accepted/0000-workspace-run-commands.md
bestander: Seems more logical to either list all packages for all workspaces or fail with error.
Picking first workspace is strange and not really useful

bestander:
@James-E-Adams, seems logical however it can be achieved with a `cd workspace && yarn run test` macro.
Because every new feature adds more support cost I would vote against it.

bestander:
@giridharangm, sorry for a delay in review.
If you are still around are you willing to go through a new review cycle?

giridharangm:
No worries @bestander . Me too, was packed with tasks (I'm a full-time dev in a private firm). I'm in the middle of a production release, but I'll update the docs as per your review ASAP.

bestander:
No rush, just ping whenever you are ready

accepted/0000-workspace-run-commands.md
ianstormtaylor: +1 to listing all of the packages in all the workspaces. That would also allow for easy "filtering" of workspaces if you only want to list a specific few of them, like:

```
yarn workspaces --filter workspace-a list
```

...or similar. (I think something like `--filter` functionality would ideally be available on all the `workspaces` commands.)

accepted/0000-workspace-run-commands.md
```diff
@@ -0,0 +1,67 @@
...
+
+## `yarn workspaces list`  
+
+This command lists out the package names inside that workspace.  In the case of multiple workspace configuration, we'll need to pass the workspace name like `yarn workspaces <workspace_name> list`. By defaut, when no workspace name is not given, it'll pick the first workspace name from the root package.json.  
+
+## `yarn workspaces run <command>`  
+
+This command will execute the specified command in all the packages under the workspace. Before we run the scripts, it should make sure that all packages has the script specified in the package.json. Otherwise, it should throw an error specifying which pacakge is missing that script.  
+
+Just like the `list` command, by default, it'll select the first workspace in case of no workspace name is provided.
```
ianstormtaylor: I think this should again run the command in all workspaces, not just the first one. And then later a "filtering" mechanism can be added to restrict further. But otherwise it seems to arbitrary.

giridharangm: I think, like @bestander said earlier, if we want to execute a command in a specific workspace, it's better to `cd` to that package and `yarn run`. 

I've updated this command to execute all in all workspaces.

accepted/0000-workspace-run-commands.md
giridharangm: On a second thought, yeah, you're right. This doesn't gives any extra convenience. Removed!

accepted/0000-workspace-run-commands.md
giridharangm: @ianstormtaylor thanks for that `--filter` tip.

giridharangm:
@bestander I've made some few change. Could you please have a look?

ianstormtaylor:
@bestander would you reconsider the ability to offer filtering to run commands in specific subsets of packages? I think there are three usage patterns for this:

1. Executing a script in all packages.
2. Executing a script in a specific subset of packages.
3. Executing a script in a single package.

I agree that (3) can be easily solved by `cd ... && yarn run ...`, but only the option of some sort of filtering/scoping mechanism (like Lerna has) allows you to solve all three easily.

My use case for (3) is having workspaces split up into `services` and `utils`, like so:

```js
workspaces: [
  'services/*',
  'utils/*',
]
```

Which allows me to keep things more organized, and then for example I could run build watchers across all workspaces, but only run server-starting commands in the `services` workspaces.

bestander:
To get the ball rolling let's merge this one.

bestander:
@ianstormtaylor, sure, can you send a PR for the same file with the proposed change?
Let's discuss it pros and cons

Siyfion:
Just to add my 2¢...

I personally think this is a great idea, and will certainly help a lot when dealing with a monorepo. I also can't think of anything that this change would "break" in current functionality, which is always good! Lerna is fantastic and all, but not all monorepo projects need/should to be published to NPM.

We're currently investigating using monorepos to hold a number of SPAs that are all for a given "project". So for CI/CD integration this would be hugely useful as well as for local development.

PR #84, opened 2017-09-21
add rfc about registry url in lock file

vernak2539:


accepted/0000-registry-url-in-lock-file.md
```diff
@@ -0,0 +1,57 @@
...
+The expected output is for yarn to recognise the registry URL supplied (via either cli flag or
+`.yarnrc` or `.npmrc`) and download from that if the registry used in the lock file is not available.
+
+
+# Detailed design
+
+Let me know if this isn't enough, first time writing an RFC.
+
+
+As a dev that generated their lock file against REGISTRY_A
```
bestander: Here we need a more technical explanation with examples of yarn.lock and config parameter names and concrete URL examples.

accepted/0000-registry-url-in-lock-file.md
```diff
@@ -0,0 +1,57 @@
...
+I would like to pass a cli flag designating REGISTRY_B
+And I would like yarn to download my dependencies from REGISTRY_B
+
+As a dev that generated their lock file against REGISTRY_A
+I would like to overwrite my `.yarnrc` or `.npmrc` designating REGISTRY_B
+And I would like yarn to download my dependencies from REGISTRY_B
+
+# How We Teach This
+
+This would largely be in the internal workings, and wouldn't require teaching. npm 5 does this
```
bestander: We just need an entry in docs page, maybe here https://yarnpkg.com/en/docs/yarnrc.

A short blog post would be great too.

It is straight forward, so no need to mention anything else here.

bestander (changes requested):
Thanks, @vernak2539  that is a great start!
Just some cleanup and more concrete examples would be enough.

accepted/0000-registry-url-in-lock-file.md
vernak2539: @bestander updated the PR. Let me know if more is needed

accepted/0000-registry-url-in-lock-file.md
```diff
@@ -0,0 +1,71 @@
...
+- Yarn Issue:
+
+# Summary
+
+The yarn.lock file includes information about the registry that it was generated against. It would
+be nice if there was no registry information associated with the lock file
+
+# Motivation
+
+We currently have two internal registries/mirrors. We generate the lock file against one we have in
```
bestander: Add more details please.
Short examples how the yarn.lock files are different would be useful

accepted/0000-registry-url-in-lock-file.md
```diff
@@ -0,0 +1,71 @@
...
+our development environment. When our code is deployed to our build/testing environment, we
+no longer have access to the original registry that was used to generate the lock file. This causes
+our download to error as yarn cannot retrieve the packages.
+
+The expected output is for yarn to recognise the registry URL supplied (via either cli flag or
+`.yarnrc` or `.npmrc`) and download from that if the registry used in the lock file is not available.
```
bestander: no need to do it in .npmrc, just mention .yarnrc.

accepted/0000-registry-url-in-lock-file.md
```diff
@@ -0,0 +1,71 @@
...
+be nice if there was no registry information associated with the lock file
+
+# Motivation
+
+We currently have two internal registries/mirrors. We generate the lock file against one we have in
+our development environment. When our code is deployed to our build/testing environment, we
+no longer have access to the original registry that was used to generate the lock file. This causes
+our download to error as yarn cannot retrieve the packages.
+
+The expected output is for yarn to recognise the registry URL supplied (via either cli flag or
```
bestander: please provide the name of the flag

accepted/0000-registry-url-in-lock-file.md
```diff
@@ -0,0 +1,71 @@
...
+
+As a dev that generated their lock file against REGISTRY_A
+I would like to pass a cli flag designating REGISTRY_B
+And I would like yarn to download my dependencies from REGISTRY_B
+
+As a dev that generated their lock file against REGISTRY_A
+I would like to designate an alternate registry URL in my `.yarnrc` or `.npmrc` (REGISTRY_B)
+And I would like yarn to download my dependencies from the alternate URL (REGISTRY_B)
+
+I feel like the simplest solution would be to add a "hash" or "commit-ish" field to entries.
```
bestander: That is a lot of text that does not add much value compared to the previous section.
Here we need to see concrete examples how Yarn will behave with the new feature.
I suggest something like this:


If yarn.lock file contains
```
abbrev@1:
   version "1.1.0"
 ```
And .yarnrc has setting `override-registry={"registry.yarn.pkg.com": "registry.npm.org"}`
Then Yarn will fetch `abbrev-1.1.0.tgz` from `https://registry.npm.org/abbrev/-/abbrev-1.1.0.tgz#d0554c2256636e2f56e7c2e5ad183f859428d81f` instead of the one specified in yarn.lock.

File: accepted/0000-registry-url-in-lock-file.md [hunk 7]
```diff
@@ -0,0 +1,71 @@
...
+```
+abbrev@1:
+  version "1.1.0"
+  hash "d0554c2256636e2f56e7c2e5ad183f859428d81f"
+  resolved "https://registry.yarnpkg.com/abbrev/-/abbrev-1.1.0.tgz#d0554c2256636e2f56e7c2e5ad183f859428d81f"
+```
+
+This, with the version already provided, could then be used to generate a URL on the fly or at the beginning of operations if an alternate registry is specified in the scenarios above.
+
+Or, the `resolved` field could change to not include the hostname/origin, which would then be inserted during operations. This would
```
bestander: Let's keep this proposal simple and not have many possible solutions in one place, you can list them below in the alternatives section

accepted/0000-registry-url-in-lock-file.md
```diff
@@ -0,0 +1,71 @@
...
+probably have implications for backwards compatibility though.
+
+
+# How We Teach This
+
+This would largely be in the internal workings, and wouldn't require teaching. npm 5 does this
+already, so it would be inherent.
+
+> If you generated your package lock against registry A, and you switch to registry B, npm will now try to install the packages from registry B, instead of A. If you want to use different registries for different packages, use scope-specific registries (npm config set @myscope:registry=https://myownregist.ry/packages/). Different registries for different unscoped packages are not supported anymore.
+
```
bestander: Then let's just copy what npm5 does.
Make a proposal to match npm5 behavior with similar examples and configs

bestander (changes requested):
Good start, @vernak2539.
I've left a few comments

accepted/0000-registry-url-in-lock-file.md
vernak2539: If we want a proposal to do just like npm, nothing would need to change, just yarn would have to take into account the registry in the rc file, and use the registry in that no matter how the lock file was generated. I think that's where I was going with the original one, but that would be a fundamental change in behaviour, probably a breaking change

csvan: Can I contribute something here? Very eager to see this go through - it is a dealbreaker at the moment for us since we (like @vernak2539 ) use different registries in different build environments.

bestander: @csvan, everyone's contribution is welcome.

bestander:
Overall I like it.
Thanks for following this up, @vernak2539, very much appreciated.
Does anyone have anything to add?

If not then let's merge it and wait for a PR.

bestander (approved):
LGTM, waiting for more comments.
If none please ping me to merge this as is in 5 days.

BYK:
I think this is a good approach but why not go one step further and remove the registry URLs from the lockfile completely?

bestander:
>  but why not go one step further and remove the registry URLs from the lockfile completely?

1. backwards/forwards compatibility
2. easier for Yarn to recognize the pattern as registry sourced
3. smaller change required

imsnif:
Hey, thanks to @csvan for pointing this out! We're feeling this pain when moving registries (https://github.com/yarnpkg/yarn/issues/4817)... This might seem trivial, but I want to make sure (since I did not see this mentioned) - will the added flag also recreate the package checksum? (the checksum ultimately includes the repository url, so if a package was re-published to another registry we'll get a checksum mismatch).

bestander:
Commented on the linked issue, I think this is a separate thing

BYK:
@bestander 

>backwards/forwards compatibility
>easier for Yarn to recognize the pattern as registry sourced

We can probably add a new field per resolution and signal this easily. That should be forward and backwards compatible. Only thing is new lock files wouldn't work with old Yarn versions but I think that was never a concern.

>smaller change required

With my suggestion above, I'm not sure if it would be that big of a difference. The advantage here is avoiding extra processing per field.

bestander:
@BYK, I am ready to discuss the change. :)
Here is my argument - making this change will be a safe and implementation could be fast.

Stripping npm/yarn domain needs way more than just one step forward, we need to think of the migration path.
I think that kind of change is worth a major version update even though coding it is easy.

bestander:
To keep the ball rolling I'll merge this as is.
Very much welcome to enhance it in future iterations but I think we are good to go to get the minimum viable solution in.

@vernak2539 are you eager to send a PR?

accepted/0000-registry-url-in-lock-file.md
Remigius2011: Hi,

our use case is sharing code between external (outside the company, seeing directly public npm repos) and internal developers (i.e. working inside a company, with proxy and an artifactory mirroring the public npm repos). moreover, there is a ci that also sees only internal repos.

actually, there are two issues with repos in yarn.lock:

1. while reading (i.e. when installing npm modules)
2. while writing (i.e. when new modules are added)

my remarks / desires (without regard whether they are fulfilled by the suggested solutions above):

for 1.: basically, what we need is to have a single yarn.lock that is valid and working in both environments, so it can reside in our git repo. even if we would be fine without them, yarn.lock must have embedded repo URLs for backward compatibility. hence we would like to have a single set of URLs, which should be: public URLs for public available npm packages and private (i.e. pointing to the artifactory repo) URLs for private packages (actually, the company has separate artifactory URLs for public and private npm packages). obviously, access to the private packages is only possible from inside the company, so this would break for external developers - which is OK.

for 2.: the desire to have a single set of URLs both valid inside and outside the company (obviously this is only possible for public packages) would mandate that public URLs are written into yarn.lock even when a dependency on a public npm package is added by an internal developer and thus served by the internal artifactory repo. note that this use case does typically not occur during a build process (e.g. in a ci/cd). of course, the package namager has no other means to decide whether a package is public than configuration and the final package URL (in our case, it would be possible to make this distinction, as private packages have a different internal URL, in other cases there might be a set of rules giving such a distinction - which in turn might be overshooting to support).

thus we would probably need means to override a certain public repo by a private one as a *local configuration property* (not going into the git repo - e.g. in .npmrc) and simultaneously using this override "backwards" when writing URLs into yarn.lock. note also that the override cannot only be only a host, as the URL path and/or the protocol under which npm modules are visible may differ between different repos.

arcanis:
Regarding the implementation, I have two comments:

- Not a fan of passing a JSON object as command line argument. Do we need this option? I feel like the most common use case will be to switch the default registry for everything at once (for example to use `registry.npmjs.org` instead of `registry.yarnpkg.com`), so I'd prefer if it was possible to do this without having to specify a full JSON object (which doesn't play well with shellscripts variables).

- The yarnrc files use our lockfile file format, which supports dicts. It would be better to use this syntax instead of encoding the registry mapping with JSON.

arcanis:
I also note the following comment quoting the npm behavior:

> Different registries for different unscoped packages are not supported anymore.

Why do we need to support a registry=>registry mapping, then?

rally25rs:
> I think this is a good approach but why not go one step further and remove the registry URLs from the lockfile completely?

Since this _could_ be slated for yarn v2, we could ignore backwards compatibility. IMO adding another setting to override the registry is just going to cause confusion between when to use `registry=` and `override-registry=`

kopax-polyconseil:
1. The checksum of the yarn lock change if you edit it, that's one problem
2. The git repo not being up to date after editing the yarn lock, a second problem. 
3. A workaround would be to allow to specifiy the lock file then create a 2nd one with the first one using the other repo, that's not possible
4. A setting stronger than `--repository` that could take precedence over the lock file, also not possible.

Any plan to implement a solution within yarn after all those years ?

accepted/0000-registry-url-in-lock-file.md
kopax-polyconseil: any update on this rfc ?

PR #85, opened 2017-10-22
move selective version resolutions to implemented

victornoel:
It was implemented, thanks to @kaylieEB :)

If there are adjustments to make following the implementation, don't hesitate!


kaylieEB:
Thanks @victornoel !

PR #86, opened 2017-11-04
nohoist RFC

connectdotz:
addressing [#3882](https://github.com/yarnpkg/yarn/issues/3882)

### DIFF
--- a/accepted/0000-nohoist.md
+++ b/accepted/0000-nohoist.md
@@ -23,13 +23,30 @@ See issue [#3882](https://github.com/yarnpkg/yarn/issues/3882) for why developer
 
 # Detailed design
 
+## Summary: what does nohoist really do
+At the end of the day, nohoist is just a mechanism to determine a different 'hoist' point. By default, **all** packages in the workspaces are hoisted to the **root**, except for version conflicts (similar to [npm resolution](https://docs.npmjs.com/how-npm-works/npm3)). Nohoist provided a mechanism to custom this algorithm and determine:
+1. what packages should be the exception of the rule ?
+
+    This is majority of the work for the rest of the RFC, we will describe how we can use the glob pattern to match the package in the dependency graph, like a virtual file system.
+
+2. where to place them?
+    There are general 2 options:
+    - option-1: under the module that reference it. 
+    - option-2: under the workspace that reference it.
+
+    I have tried both implementations and realized why most package manager moving to the hoisting model... By leaving the packages under the referencing module (option-1), I quickly bumped into the scalability issue (ran out of ram during `yarn install`) for larger project due to high number of (duplicated) modules. Not to mention the complexity in resolving the [modules-in-between](#module-in-between) issue...
+
+    After examined how react-native project worked outside of the workspace environment (both yarn and npm did host everything to the \$cwd/node\_modules), I realized that developers are already trained to go **down** the node\_module tree, but not every one has followed the [node module resolution algorithm](https://nodejs.org/api/modules.html#modules_loading_from_node_modules_folders) to go **up** from \$cwd. It is generally not a problem, unless you are in the monorepo project where modules can reside above the workspace referencing them...
+
+    This drives me to settle on option-2 by simply hoisting the nohoist modules in the workspace's node\_modules (instead of the project's root) where they can still share the benefit of hoisting and greatly reduce the complexity in implementation and education. This essentially made the workspace/node_modules looked like a stand alone project, for the given nohoist-modules. Therefore, if a package can work in a stand alone project, it should be able to work under workspaces.
+
 ## principle
 - hoist as much as we can
 - nohoist is an exception/workaround, therefore favor 
     - explicitness over convenience
     - small scale over large scale
 
-These principles lead us to the "**shallow nohoist**" implementation: _nohoist only apply to the package explicitly specified; all of its dependency, unless also on the nohoist list, will be hoisted by default._
+These principles lead us to the following implementation: _nohoist only apply to the package explicitly specified; all of its dependency, unless also matched the nohoist list, will be hoisted by default._
 
 ## divide the problem 
 
@@ -43,6 +60,7 @@ hoist-all(1) --> nohoist-direct-dep(2) --> nohoist-deep-dep(3) --> hoist-none(4)
 4. hoist-none: a.k.a nohoist all, opt-out hoisting all together: use-case #3.
 
 yarn currently support (1), this RFC proposed solution to handle (2), (3), (4).
+
 
 ## configuration
 We need a place to specify "nohoist" packages for the given workspace(s). We can expand the existing [workspaces](https://yarnpkg.com/en/docs/workspaces) config:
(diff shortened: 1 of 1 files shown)


bestander:
Conceptually seems legit and reasonable.
Thanks for sending a very well thought through RFC, @connectdotz!

@BYK and @arcanis I'll leave the details review to you.

connectdotz:
@bestander  I just updated the RFC to callout nohoist exceptions for linked packages (including workspace itself): basically linked packages will always be hoisted. Please let me know if you see any issue with that.

Not sure what the expected end result for an RFC... do we need explicit approval before submitting the code change PR? I assume the state of the RFC is still pending on @BYK and @arcanis review?

accepted/0000-nohoist.md
```diff
@@ -0,0 +1,299 @@
...
+    workspaces?: Array<string> | WorkspacesConfig
+}
+```
+note: the config will be backward compatible.
+
+### nohoist list inheritance
+nohoist config is inherited, i.e. if nohoist is specified in the root package.json, all workspaces will inherit the nohoist list, likewise all the dependencies of the workspace will inherit the nohoist list. This concept is simple and useful not only in reasoning nohoist from root to workspaces, it can be extended downwards to allow packages provide better monorepo support as well as easily handling deep-dependency as direct-dependency. 
+
+### nohoist list naming
+lerna uses glob pattern to specify nohoist package, however based on the design principle, I argue it is probably better not to since nohoist should be "rare", therefore probably doesn't justify for extra matching complexity and risk.
```
bestander: nice approach!
Simplicity is important

accepted/0000-nohoist.md
```diff
@@ -0,0 +1,299 @@
...
+```
+note: the config will be backward compatible.
+
+### nohoist list inheritance
+nohoist config is inherited, i.e. if nohoist is specified in the root package.json, all workspaces will inherit the nohoist list, likewise all the dependencies of the workspace will inherit the nohoist list. This concept is simple and useful not only in reasoning nohoist from root to workspaces, it can be extended downwards to allow packages provide better monorepo support as well as easily handling deep-dependency as direct-dependency. 
+
+### nohoist list naming
+lerna uses glob pattern to specify nohoist package, however based on the design principle, I argue it is probably better not to since nohoist should be "rare", therefore probably doesn't justify for extra matching complexity and risk. 
+
+However, it is not feasible to ask developers mark off every package if they just want to opt-out hoisting all together, therefore a simple keyword like "\_all\_" should be introduced to handle use-case-3 easily. For example the following config will basically turn off nohoist for the whole monorepo project:
```
bestander: I am not a huge fan of a new keyword.
Would it really be simpler than a glob?

Considering that we use globs already I think it would be expected by final users that nohoist fields is a glob too

accepted/0000-nohoist.md
```diff
@@ -0,0 +1,299 @@
...
+Let's say a user project A has a dependency on a public package B, which has a dependency on 'react-native' that will need to be excluded from hoist otherwise it won't work. 
+
+||B has workspaces.nohoist=react-native|B didn't have workspaces.nohoist|
+|--|--|--|
+|A doesn't use yarn or workspace|no impact|no impact|
+|A uses yarn workspace|react-native will be excluded from hoisting without A doing anything|A needs to discover then put both B and react-native in its own workspaces.nohoist|
+|A uses yarn workspaces but excluded \_all\_|no impact, none will be hoisted|no impact, none will be hoisted|
+|A uses yarn workspaces but do not want to exclude react-native from hoisting (why?)|conflict! but the public package B is right and react-native should be excluded.|no conflict but B will most likely failed|
+
+In short, having unnecessary nohoist packages might cause inefficiency, but missing a necessary nohoist will lead to compile/execution error.
```
bestander: That is a good argument.
The counter argument is if we release it now and it goes to public packages then we won't have a chance to change our minds.
Especially if other package managers will implement workspaces.

How about this - we implement it the safer way, `nohoist` only in root but we can implement the nohoist in public packages behind an experimental flag name  `nohoist-yarn-experimental`.
Give it some time and then decide how to proceed.

accepted/0000-nohoist.md
```diff
@@ -0,0 +1,299 @@
...
+```
+- option-2: if we don't remove the nohoist list from A, package B will stay under A as it is in the nohoist list:
+```
+workspace-a
+A -> B
+```
+IMHO, option-2 is more confusing than option-1. We can reason the nohoist inheritance is based on the dependency tree, once the tree is broken (by hoisting), so should the nohoist list inheritance. But I am open for better arguments.
+
+## unit tests failed on mac
+this is not related to nohoist issue specifically, but made submitting PR much more difficult and time consuming. Many async integration tests failed on my laptop (mac) for even a fresh clone. Is this a known issue? any suggestion or workaround?
```
bestander: Nope, unit tests should all pass.
Submit a PR and let's see what CI says

accepted/0000-nohoist.md
```diff
@@ -0,0 +1,299 @@
...
+workspace-a
+A
+B
+```
+- option-2: if we don't remove the nohoist list from A, package B will stay under A as it is in the nohoist list:
+```
+workspace-a
+A -> B
+```
+IMHO, option-2 is more confusing than option-1. We can reason the nohoist inheritance is based on the dependency tree, once the tree is broken (by hoisting), so should the nohoist list inheritance. But I am open for better arguments.
```
bestander: I understand that `nohoist` only works for workspace boundary, regular packages should not inherit workspaces' logic for hoisting.
So yeah, option-1 looks the right one

accepted/0000-nohoist.md
connectdotz: @bestander I see your point, yes we can indeed use glob pattern to make it consistent with the packages patterns. 

BTW, I loved the "why" feature of yarn, however, while using it to diagnose nohoist, there are a few bugs surfaced, mainly the current implementation assumed a given package only appeared once, which is not always true: 1. when version conflict 2. when it is marked nohoist. I will address them also in this PR so people can easily find out why a package appeared in multiple places.

(There is another why.js bug that it didn't look at peerDependency, but I will not include it in this PR, as it is independent of nohoist, trying to keep the scope tight.)

accepted/0000-nohoist.md
connectdotz: ok, in that case, we can just use the existing `workspaces-experimental` flag, since nohoist is within the workspaces config anyway, probably don't need another flag for nohoist explicitly.

Basically, both package producers and consumers need to have  `workspaces-experimental` flag on in order to access any workspaces config/functionality.

accepted/0000-nohoist.md
connectdotz: @bestander 
> regular packages should not inherit workspaces' logic for hoisting

If by `regular packages` you mean the project without turning on `workspaces-experimental`, then yes nohoist only impacts the hoisting logic via `workspaces.nohoist` in package.json.
 
The more I think about the original question raised in the RFC, the more I realize that it has to be option-1. option-2 is not a valid alternative actually. Because a hoisted package can be "originated" from multiple sources that might have conflicting nohoist rules, therefore, inheritance-from-source is problematic. 

This also helps us simplify nohoist-inheritance concept, as it simply traverses the same node_module graph, just like package resolution. 

I will update the RFC to reflect this.

brunolemos:
Is this available as a beta or something like that?
Need to make react native command work.

connectdotz:
quick update: 

last week was a busy one but I managed to spend the weekend on this. Finally able to use the new nohoist for a fresh react-native run in my sandbox workspaces... Bumped into quite a few incompatible hidden assumptions in the non-hoisting part of the system and discovered that we do need deep-nohoist for react-native (Metro bundler etc)... anyway, I should be able to submit the PR this week so we can have more concrete discussions and more people testing it...

had to change the design a bit to provide a more powerful/flexible deep-nohoist mechanism. Fortunately, @bestander suggestion of using glob pattern matching provided an intuitive way to do this. I will need to update this RFC as well...

accepted/0000-nohoist.md
```diff
@@ -252,13 +253,22 @@ the file structure will look like this:
...
-    - More implementation alternatives are also marked as discussion in their corresponding sections above.  
-    - we could choose not support nohoist-deep-dep(3), as use case (#2) is not very clear to me if we need this or what the expectation is. However, with inherited nohoistList, we get deep-nohoist pretty much for free so decided to just leave it in. We could either not document it publicly or leave it as it is and refine later when needed.
+    - Some implementation alternatives are discussed inline above. 
+    - The rest is documented in the following section.
 
 # Unresolved questions
 
-## discussion-1
-**is allowing nohoist in public package bad?**
+## dependency tree representation
```
connectdotz: @bestander @BYK @arcanis would love to hear your thought on this issue...

bestander: Thanks, I'll check this later today or tomorrow

arcanis:
@connectdotz Great job on this RFC! My feedback:

- A separate flag from `workspaces-experimental` will be needed, since this flag now defaults to `true` and will be removed in a few releases from now.

- I feel strongly against allowing the nohoist option in non-private packages (they don't need to be workspace roots, but they need to be marked as `private: true`). Doing this would remove users the control over the specific of their package managers.

- As for the dependency tree representation, we should use the same one than the one used for the [selective version resolution feature](https://yarnpkg.com/en/docs/selective-version-resolutions#toc-how-to-use-it). Which is great, since it's very similar to the syntax you came up with :)

connectdotz:
@bestander @arcanis thx for the feedback, addressing @arcanis comments:

> A separate flag from workspaces-experimental will be needed, since this flag now defaults to true and will be removed in a few releases from now.

I see, then it does make sense to have a new one. I will add `workspaces-nohoist-experimental` opt-out option.

>  feel strongly against allowing the nohoist option in non-private packages (they don't need to be workspace roots, but they need to be marked as private: true). Doing this would remove users the control over the configuration of their package managers.

I wasn't quite sure what you refer to as "_Doing this would remove users the control over the configuration of their package managers._"  Each package, public or private, is indeed _giving package manager instruction_ on what to load. It is true that they usually don't care where it should be loaded, but for some cases that the position of the module is critical like the react-native module,  somebody has to specify the exception. We can ask each consumer (i.e. our projects) to mark the exception, or the package owner can mark the exception itself. I think the later is more efficient and scalable.

If you are concerning malicious "take over", I should have documented it better that nohoist rule is always prefixed with the "dependency path" at runtime, so they can really only influence their own packages tree down. For example: in a A/B/C tree, even if B has a [**"] nohoist rule, it will only impact any package match `A/B/**`, so none of A's other dependency will be impacted by B, i.e. B can only impact its own dependencies' hoisting strategy.

If for whatever reason we need to turn off a public package's nohoist rule, the workspace can always use glob's negation operation to override package's. (I didn't implement this in the initial PR as it can make the nohoist syntax a bit messy, but we can easily enable it later, if need to.)

Can a public package do stupid thing? Yes, they could do many stupid things like adding unnecessary dependencies or adding unnecessary nohoist rules. They can cause our workspace less efficient but will most likely not break it. However, if the package didn't specify a needed nohoist rule, our workspace will not work. So my thinking goes like this: 1) there are reasonable protections to sandbox the nohoist rules to its originating package or override it; 2) abusing the rule causing less harm than not being able to specify it... therefore, nohoist for public package should be considered and even encouraged if the package requires it to work...;-)

> As for the dependency tree representation, we should use the same one than the one used for the selective version resolution feature. Which is great, since it's very similar to the syntax you came up with :)

Just to clarify... you are saying 'yes we should use path-like syntax to express dependency tree', right? I think we all agree that for configuration, glob pattern is the one to use, such as in workspaces.packages, version-selection and now nohoist. 

Sorry for not being clear, what I meant to ask is specifically regarding reporting progress and state, such as during install or why. For example when asking `yarn why is-finite` in a nohoist react-native workspace, it (PR [#4979](https://github.com/yarnpkg/yarn/pull/4979)) will produce the following result:
```
$ yarn why is-finite
yarn why v1.3.2
[1/4] 🤔  Why do we have the module "is-finite"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
[4/4] 🚡  Calculating file sizes...
=> Found "is-finite@1.0.2"
info Reasons this module exists
   - "/_project_/ites/ts-jest/babel-core/babel-generator/detect-indent/repeating" depends on it
   - Hoisted from "repeating#is-finite"
info Disk size without dependencies: "16kB"
info Disk size with unique dependencies: "32kB"
info Disk size with transitive dependencies: "32kB"
info Number of shared dependencies: 1
=> Found "ites#is-finite@1.0.2"
info Reasons this module exists
   - "/_project_/ites/react-native/metro-bundler/babel-generator/detect-indent/repeating" depends on it
   - Hoisted from "ites#repeating#is-finite"
   - in the nohoist list ["/_project_/**/react-native","/_project_/**/react-native/**"]
info Disk size without dependencies: "16kB"
info Disk size with unique dependencies: "32kB"
info Disk size with transitive dependencies: "32kB"
info Number of shared dependencies: 1
✨  Done in 1.71s.
```

You can see there are mixed dependency tree expressions. The new ones use standard path-like format('/'), the rest use yarn custom format ('#'). We should be consistent, if possible. IMHO, if there is no particular reason in choosing '#', we should probably consider standardizing on the path-like format instead...

arcanis:
> Each package, public or private, is indeed giving package manager instruction on what to load.

Yes, but they don't give the package manager instructions on how to store it on the filesystem. Two actual examples:

- The dependencies lockfiles are ignored - only the top-level one counts
- There is no way for a dependency to request to be installed in flat mode

> We can ask each consumer (i.e. our projects) to mark the exception, or the package owner can mark the exception itself. I think the later is more efficient and scalable.

I see your point, but they're other factors into play:

- Adding this feature to the package level encourages package developer to rely on those mechanisms, which are broken imo (in that they prevent the package manager from actually doing its job efficiently). Requiring this feature to be described in the top-level package.json makes it clearer that relying on the hoisting is not a pattern we want you to use, while unblocking the few packages that actually need it.

- Adding this feature to the package would be simultaneously too much and not enough: too much because it would silently affects the file tree in quite unpredictable ways, and not enough in the sense that if a package makes assumption on the hoisting of its dependencies, it also makes sense to think that it will also make such assumptions on the hoisting of its siblings (which it wouldn't be able to `nohoist`, per the rule you detailed above).

I'm not so concerned about malicious packages, however I'm concerned about the ecosystem, and I prefer not to encourage a pattern that effectively decreases the user experience.

> We should be consistent, if possible. IMHO, if there is no particular reason in choosing '#', we should probably consider standardizing on the path-like format instead...

Oh, I see. Yes, I agree it would be better to be consistent. However, I think this is worth being discussed in a separate PR - I'd suggest to use `#` for the time being and to revisit the output format in a followup issues.

connectdotz:
@arcanis I think we might have misaligned on a very important assumption: nohoist is not a stand-alone 'feature' for packages to take-over location decision from PM, it is merely a 'workaround' to make an otherwise 'working' package function within the monorepo project. nohoist is an no-op outside of workspaces. The developers have to make their packages work without nohoist (most of their consumers will not in monorepo env), and only add nohoist if some of its dependency hasn't been updated to work in monorepo env (we can easily add a warning to make it clear that such package should be fixed soon). In other words, There is no incentive for developers to add unnecessary nohoist.

If this is still not enough to make you feel comfortable, then let's just make it private only as you suggested😉 . Since many workspaces are also public packages, this change will mean that only the root package.json can specify nohoist. I think it is probably all right for now.  We are just debating on some hypothetical situations that we might not yet fully understand anyway... If we ever change our mind, it is literally a 1-line code change...

> I'd suggest to use # for the time being and to revisit the output format in a followup issues.

hmmm... ok

arcanis:
> There is no incentive for developers to add unnecessary nohoist.

Eh eh, I'm afraid if it worked like that no package would have ever relied on the hoisting mechanism in the first place :p

Ok, let's do the private thing, and I'll go review your implementation PRs!

connectdotz:
@bestander @arcanis code changes and RFC all updated per our discussion

PR #87, opened 2018-01-31
update workspace-run-commands to more detail

ianstormtaylor:
cc @bestander from discussion in #83 

This expands the existing `workspace-run-commands` to add more detail, and borrow a few more of the useful pieces of the functionality Lerna offers for running commands in workspaces.

Specifically, it introduces:

- Executing arbitrary commands, not just package scripts with `yarn workspaces exec`.
- Passing extra arguments to the package scripts with `yarn workspaces run`.
- Controlling whether to run in parallel or not with `--parallel`.
- Controlling the concurrency with `--concurrency`.
- Filtering the commands to a specific set of workspaces with the `--workspaces` flag.
- Filtering the commands to a specific set of packages with the `--packages` flag.

Almost all of this functionality is based on Lerna's prior art, which has been super useful to managing monorepos in my experience—for all things from running tests, to building source, to automating other useful tasks.

---

One open question I have is how the Yarn team thinks about what a "workspace" is. In the `package.json` configuration, they're defined like:

```json
"workspaces": [
  "packages/*",
  "services/*",
  "utils/*"
]
```

Which makes it simple to have multiple directories of workspaces, or to list individual top-level directories as workspaces too. This is super useful in my experience.

But it also calls into question how we should be referring to specific folders inside those directories. Is the `packages` folder a "workspace"? Or is each of the directories inside `packages` a "workspace"?

This question becomes more important when trying to figure out how to let people filter things. Because it would be super useful to be able to do something like:

```
$ yarn workspaces run --filter 'services/*' test
```

To restrict the command to run only in the `services/*` workspaces above.

Lerna solves this by instead calling the configuration `packages`, and then only ever refers to packages by their name in `package.json` instead, which simplifies things. It's possible that Yarn should follow their lead, and ignore the idea of a "workspace name", and only treat package names. (Which would eliminate the `--workspaces` flag in this pull request.)

In this sense, the `--packages` and `--workspaces` flags are just strawmen. If there were to only be a single filtering mechanism, it could use `--only` and `--ignore` flags instead.

That said, it feels slightly weird for Yarn to declare workspaces the way it does, and then not to let people filter by `services/*` or `packages/*` in this case. Thoughts?

### DIFF
--- a/accepted/0000-workspace-run-commands.md
+++ b/accepted/0000-workspace-run-commands.md
@@ -4,53 +4,128 @@
 
 # Summary
 
-Allow Yarn CLI to execute a package script on each packages from the workspace root.
+Allow Yarn CLI to execute a package script on each workspace package from the workspace root.
 
 # Motivation
 
-"Lerna". Lerna does a great job for handling monorepos. Since Yarn has a built-in
-workspaces feature, we could use some of the functionalities that being used in
-lerna (not entirely though!) for handling the monorepo more efficiently.
+Lerna does a great job for handling monorepos. Yarn's built-in workspaces feature would benefit from borrowing more of functionality that Lerna exposes, to handle monorepos more easily.
 
-Just like installing all dependencies from one place (workspace root), it'd be great
-to execute all the sub-package scripts from the workspace root using a single command.
-Not only this would make a less back-and-forth traveling between the packages to
-execute scripts, but also it'll help a lot for CI/CD configuration.
+Just like installing all dependencies from one place (workspace root), it'd be useful to execute a specific package script for each of the workspace packages from the workspace root using a single command. Not only this would make for less back-and-forth traveling between the packages to execute scripts, but also it'll help a lot for CI/CD configuration.
 
 # Detailed design
 
-As @BYK mentioned [here](https://github.com/yarnpkg/yarn/issues/4467#issuecomment-330873337),
-we could create two commands for this feature.
+As @BYK mentioned [here](https://github.com/yarnpkg/yarn/issues/4467#issuecomment-330873337), to start we'd create two commands for this feature:
 
 * `yarn workspaces list`
 * `yarn workspaces run <command>`
 
-By introducing `workpaces` command, it brings a modular approach from the CLI's perspective.
+Unifying all of the workspace-specific commands under the `workpaces` namespace allows for a modular approach from the CLI's perspective, making future commands easy to add.
 
-## `yarn workspaces list --args`
+## `yarn workspaces list [flags]`
 
-This command will list out all the packages for all the workspaces (alphabetically). If no workspace is found, this will fail with error.
+This command lists all of the packages for all of the workspaces alphabetically. If no workspaces are found it will error.
 
-Additionally, we can pass a `--filter` argument followed by a workspace name. In that case, it'll list out all the package inside that specified workspace.
+## `yarn workspaces run [flags] <command> ...`
 
-## `yarn workspaces run <command>`
+This command runs a specific package script (as defined in the `scripts` property in `package.json`) for all of the packages for all of the workspaces.
 
-This will execute the specified `<command>` in all workspaces. For a _fail fast_ operation, we could traverse all the workspaces to check whether the specified command exists within that workspace, before we actually start executing. If not found, display an error with which workspace is missing that command.
+Just like the `yarn run` command, any arguments after the command name will be passed as arguments to the package script.
 
-The ordering of execution is also important. It must be executed _topologically_. So that it won't break the inter-dependant workspaces.
+For a _fail fast_ operation, we could traverse all the workspaces to check whether the specified command exists within that workspace, before we actually start executing. If not found, display an error with which workspace is missing that command.
+
+The ordering of execution is also important. It must be executed _topologically_, so that it won't break the inter-dependant workspaces. From the Lerna documentation:
+
+> By default, all tasks execute on packages in topologically sorted order as to respect the dependency relationships of the packages in question. Cycles are broken on a best-effort basis in a way not guaranteed to be consistent across Lerna invocations.
+> 
+> Topological sorting can cause concurrency bottlenecks if there are a small number of packages with many dependents or if some packages take a disproportionately long time to execute. The --no-sort option disables sorting, instead executing tasks in an arbitrary order with maximum concurrency.
+> 
+> This option can also help if you run multiple "watch" commands. Since lerna run will execute commands in topologically sorted order, it can end up waiting for a command before moving on. This will block execution when you run "watch" commands, since they typically never end. An example of a "watch" command is running babel with the --watch CLI flag.
+
+The output should be prefixed with the name of the workspace (eg. `packages/package-a:`), so the user has a better idea of what is currently running.
+
+### `--concurrency`
+
+The `--concurrency` flag changes the number of child processes that are spawn when commands are run in parallel, defaulting to `4` (like Lerna).
+
+### `--parallel`
+
+If the `--parallel` flag is passed, it will runs the commands in parallel in separate child processes, instead of running them in series. This command will ignore the concurrency flag and topological sorting requirements. (Just like Lerna.)
+
+## `yarn workspaces exec ...`
+
+This commands runs an arbitrary shell command in each package. It is similar to `yarn workspaces run`, and respects the same flags, but instead of running a package script defined in `package.json` you can pass it arbitrary shell commands.
+
+This is helpful for cases where you want to execute something that isn't worthy of storing in `package.json`, often when debugging or running a one-off.
+
+For example:
+
+```
+$ yarn workspaces exec babel --out-dir ./lib ./src
+```
+
+## Common Flags
+
+### `--packages`
+
+The `--packages` flag takes a package name or glob, and it restricts the command being run to only take affect in those packages. 
+
+**Note:** this flag operates on the package names, as defined by the `name` field in `package.json` files.
+
+For example:
+
+```
+$ yarn workspaces list --packages 'babel-*' build
+```
+
+
+### `--workspaces`
+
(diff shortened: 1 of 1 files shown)


rally25rs:
Overall I like the idea. Monorepos are becoming more popular so increasing the tooling around them makes sense. I also don't like that currently I still have a `lerna` dependency just so I can `publish` in a sane way.

Can we include doing actual docs around workspaces to this RFC? The complete lack of docs is pretty frustrating. Even as a contributor I barely understand how they work. The existing `yarn workspace` command seems to take a "workspace name" but what is that? How do you name workspaces? The docs website does not have a page for that command.

Also in `src/commands/` there is a `workspace` and a `workspaces`. What's the difference?

```
~/Projects/yarn-test 🐒   yarndev workspaces
yarn workspaces v1.4.1
error Invalid subcommand. Try "info"
info Visit https://yarnpkg.com/en/docs/cli/workspaces for documentation about this command.

~/Projects/yarn-test 🐒   yarndev workspace
yarn workspace v1.4.1
error Missing workspace name.
info Visit https://yarnpkg.com/en/docs/cli/workspace for documentation about this command.
```

bestander:
If no one objects let's merge this PR next week

bestander:
I agree with @rally25rs, we need a champion to document Workspaces docs.

bestander:
@giridharangm, do you agree with the changes?

@giridharangm, @ianstormtaylor are you ready to split the work and send PRs for those features?

ianstormtaylor:
@bestander I realize this is not as helpful, but unfortunately I won't have the time to be able to submit a PR implementing this. I was just trying to be helpful by fleshing out the spec for the commands more, ensuring that we port some of the good aspects of Lerna, so that whoever does work on them has something more concrete to work from.

Btw, another open question I think during implementation will be:

- If you pass something like `--parallel/concurrent` to `yarn workspaces run` are those flags masked from the scripts underneath? Because the regular `yarn run` doesn't have flags I think, which allows it to be unambiguous in what's passed directly to further scripts.

bestander:
@ianstormtaylor, whatever help you can offer is appreciated.

> If you pass something like --parallel/concurrent to yarn workspaces run are those flags masked from the scripts underneath? Because the regular yarn run doesn't have flags I think, which allows it to be unambiguous in what's passed directly to further scripts.

Good question.
I think you can pass through arguments with `--` separator `yarn run <command> --verbose -- -command-specific-arg`.
I know a few people are not happy about this syntax and it is open to discussion

ianstormtaylor:
Ah okay cool, I didn't realize there were also flags on the regular `yarn run`, so it's not introducing any new issues, that's good!

giridharangm:
@ianstormtaylor thanks for the refinements. It looks great!. 

@bestander as much as I'd like to contribute, unfortunately, my hands are already full with multiple projects (both prod and dev) now.  Besides, I don't have that much knowledge about yarn's code base and would require me to start from the ground up 😞 . I'm awfully sorry...

bestander:
Ok, let's merge it and look for champions

donaldpipowitch:
Thank you all for the work. Just a short question: This RFC enforces that the script name is available in all packages unlike Lerna, right? (E.g. Lerna silently skips commands which aren't configured in the `scripts` section of a `package.json` which I quite like.)

giridharangm:
@donaldpipowitch Yes. it'll throw an error if the specified script is missing from any one of the packages.

ianstormtaylor:
Skipping does sound like it could be useful to me too. @donaldpipowitch do you have a specific use case in mind?

accepted/0000-workspace-run-commands.md
```diff
@@ -4,53 +4,128 @@
...
 
 # Drawbacks
 
-Complexity. Especially while executing the package scripts. We need to keep track of the package dependencies. And also, there's already a library (lerna) does the same thing and is compatible with yarn without any extra configuration.
+This will increase the complexity of Yarn, instead of letting it live in Lerna and work in tandem. Especially while executing the package scripts, because we need to keep track of the package dependencies. 
 
 # Alternatives
 
-Don't have an alternative way (I can think of) inside yarn.
+There's no current alternative using Yarn. You have to use Lerna and Yarn in concert, which causes confusion as to which one to use when.
```
spion: We made https://github.com/whoeverest/wsrun to address this, so I suppose that counts as an alternative.

spion:
Hi! Sorry about the duplicate comment - just thought I'd mention (since the RFC does not list alternatives) this alternative:

https://github.com/whoeverest/wsrun

which was made specifically with yarn in mind.

giridharangm:
@ianstormtaylor IMHO, skipping would make unintentional side effects. I think it's better to let the user know that script doesn't exists in that package rather than silently (with a subtle warning) skips it.  

In a CI pipeline, what if the user forgot to add the `test` script (it's rare, but just in case) to one of the packages? Even though if we throw a warning (silent) for missing script so that the CI would fail, I guess its more than just a "warning". It should be treated an "error".   

Clearly, you guys are the pro(s) around here. So, what do you think?

thejameskyle:
In Bolt we went with the flags `--ignore/only [name glob]` and `--ignore-fs/only-fs [path glob]` which covers everything and are really obvious names

ianstormtaylor:
@thejameskyle that sounds like a good way to go to me! Good point that the "workspace names" are always going to be valid paths anyways. I'd suggest using `--ignore-path/only-path` instead of `fs`.

(I almost submitted this comment asking what does `fs` stand for there before figuring it out 😄 )

thejameskyle:
Yeah I'd be happy with `--ignore-path/only-path` as well.

We already have implementations for many of the pieces of this btw:

- [filterWorkspaces(workspaces, opts)](https://github.com/boltpkg/bolt/blob/c9bfe29cc8cc9417a1db3d0c0f86714db4bfadc7/src/Project.js#L177-L207)
- [runWorkspaceTasks(workspaces, task)](https://github.com/boltpkg/bolt/blob/c9bfe29cc8cc9417a1db3d0c0f86714db4bfadc7/src/Project.js#L144-L169)
- [task-graph-runner](https://github.com/thejameskyle/task-graph-runner)

ianstormtaylor:
@thejameskyle looks really good!

donaldpipowitch:
> Skipping does sound like it could be useful to me too. @donaldpipowitch do you have a specific use case in mind?

It seems to become a habit while working with Lerna that many of our packages don't share some commands in all cases. Say we develop some React components in one monorepo and we add several examples as standalone applications showcasing these components, than the components maybe have a `unit` command to run unit tests which the examples don't (necessarily) have and the examples maybe have a `serve` command, because they have a `index.html` in their root which can be served which the components don't have.

I'd be fine with some additional param like `--skip-missing`/`--ignore-missing` to opt-in into this behavior, so I wouldn't need to add additional filtering with an additional `--ignore`/`--only` pattern. (That would be a different use case for: `--ignore` some package which _has_ this command, but shouldn't be executed.)

jonaskello:
Both [lerna run](https://github.com/lerna/lerna#run) and [oao run-script](https://github.com/guigrpa/oao#oao-run-script-script) seems to be doing fine by just skipping missing scripts. This seems to be the community standard, so if yarn will not follow that, I think there should be a really good reason. I think the deviation from de facto standard should be motivated with a real-world scenario where skipping is really bad.  I cannot think of one myself, but I can think of many real-world scenarios where skipping is the wanted behaviour, and I think that is why both lerna and oao went that route.

If skipping is the more commonly wanted behaviour (which it seems to be given both lerna and oao does that) then I would propose that verifying that all packages have a script is the less common scenario and therefore it would be better to have a switch like `--all` that can be added when you want to be 100% sure that all packages have the script. The `--all` switch would error if a package is missing the script.

donaldpipowitch:
Short question: Was it ever discussed alongside with this RFC to include meta data about the called processes/scripts as environment variables?

Say you call `yarn workspaces run watch` in a workspace-based project with three packages and all three packages need to listen to a unique port. You could use some "find open port"-lib for this, but because of data races this can become quite flaky in my experiences, when `watch` is called parallel. It would probably be less error prone, if we could access an environment variable like `YARN_WORKSPACES_RUN_INDEX` which has the values 0, 1 and 2 to indicate that the script was called nth-times by yarn in this case and we could use this value to define a port in the `watch` script like this: `const port = 8080 + parseInt(YARN_WORKSPACES_RUN_INDEX || 0, 10)`.

lukebatchelor:
👋 
I'm looking at spiking some of this, specifically the `--workspaces` and `--packages` flags. Just wanted to double check if those are the final decision, or whether the `--only`, `--onlyFs`, `--ignore`, `--ignoreFs` from `bolt` might be considered (I actually [implemented them](https://github.com/boltpkg/bolt/pull/25)).

Happy to go with either though, `--workspaces` and `--packages` *should* be flexible enough for what we need, but I find bolts API easier to work with (no negating globs).

Also, I see that it was suggested to add `yarn workspaces exec` as well. This is actually another thing we need in our project, but I actually found that it's possible right now with a (possibly unintended, definitely undocumented) work around.

Yarns `exec` command can run directly in the workspaces

```
$ yarn workspaces run exec pwd

yarn workspaces v1.12.3
yarn exec v1.12.3
/Users/lbatchelor/src/yarn-workspaces-testing/packages/bar
✨  Done in 0.07s.
yarn exec v1.12.3
/Users/lbatchelor/src/yarn-workspaces-testing/packages/baz
✨  Done in 0.07s.
yarn exec v1.12.3
/Users/lbatchelor/src/yarn-workspaces-testing/packages/foo
✨  Done in 0.07s.
✨  Done in 1.33s.
```

The noise can be cleaned up pretty easily too

```
$ yarn -s workspaces run exec -s pwd
/Users/lbatchelor/src/yarn-workspaces-testing/packages/bar
/Users/lbatchelor/src/yarn-workspaces-testing/packages/baz
/Users/lbatchelor/src/yarn-workspaces-testing/packages/foo
```

Just wanted to check if that was intended. I'd be happy to document it if so.

PR #88, opened 2018-02-02
Creates proposal 'needs-install'

brigand:
The proposal is very rough. Mostly looking to see how people feel about it based on their own experiences and how you'd improve the situation yourself.

This seems to manifest in different ways in different situations, including frequency of merges, frequency of dependency changes, occasional work vs consistent work, and if `yarn` needs to be potentially run in multiple directories (maybe workspaces invalidate the last one, just learning about them now).


bestander:
`yarn check` command is designed to check for consistency between package.json / yarn.lock and node_modules.
Could the problem be in documentation and we need to promote this command more?

PR #90, opened 2018-02-12
Isolated workspaces

bdwain:
@bestander this is based on discussion in https://github.com/yarnpkg/yarn/issues/4099 and would likely address https://github.com/yarnpkg/yarn/issues/4207 and  possible https://github.com/yarnpkg/yarn/issues/4219 as well.

[rendered view](https://github.com/bdwain/rfcs/blob/master/accepted/0000-isolated-workspaces.md)

### DIFF
--- a/accepted/0000-isolated-workspaces.md
+++ b/accepted/0000-isolated-workspaces.md
@@ -1,0 +1,147 @@
+- Start Date: 2018-02-12
+- RFC PR: (leave this empty)
+- Yarn Issue: (leave this empty)
+
+# Summary
+
+Working on a workspace in isolation should be as easy as working on all workspaces together. Running a yarn command in
+a workspace should behave the same whether it is part of a larger project or whether it is its own separate repo. There
+should be a option on relevant commands, `--isolated` to have a command affect a single workspace instead of all workspaces.
+
+# Motivation
+
+Currently, yarn workspaces are optimized for people making changes across multiple projects in the 
+monorepo, but they can make things more difficult if you want to work on a single workspace in isolation. While
+the automatic linking and auto-installation of dependencies for all workspaces can be convenient if you are actually working
+on all workspaces, having to build and install everything locally for multiple projects when you only work on one
+at a time can be time consuming compared to just pulling those packages down from a registry.
+
+If you could work on a single workspace in isolation as easily as if it were its own repo, the transition to a monorepo
+from separate repos would be seamless for people who focus on a single workspace/repo at a time, and one of the
+biggest reasons not to use workspaces would be removed. 
+
+# Detailed design
+
+Terminology used for this doc:
+- rootDir: the root directory of a workspaces project. the one with the package.json with workspaces in it
+- workspaceDir: Directory for an individual workspace/package
+- isolated flag: `yarn foo --isolated`
+
+In order to be able to interact with a single workspace in both isolation and in the context of the project as a whole,
+relevant yarn commands should be able to be run with an `--isolated` flag.
+
+Relevant commands (based on current functionality) include
+- install
+- upgrade
+- upgrade-interactive
+- outdated
+- check
+- list
+- test (more relevant to [run commands RFC](https://github.com/yarnpkg/rfcs/blob/master/accepted/0000-workspace-run-commands.md). see below)
+- run (more relevant to [run commands RFC](https://github.com/yarnpkg/rfcs/blob/master/accepted/0000-workspace-run-commands.md). see below)
+
+This list may grow in the future (for example if yarn adds the ability to publish all workspaces at once).
+
+If you run a command with the isolated flag and are in a workspaceDir, it will run that command as if the workspaceDir you were currently in existed in isolation (with few exceptions detailed below).
+
+If you run a command with the isolated flag but you are at the rootDir (or any other directory not underneath a workspaceDir), it will give an error saying that the isolated flag can only be used inside an individual workspace.
+
+## Command behavior
+### install
+#### With isolated flag
+Install only that package's dependencies, and do not hoist them to the rootDir. Install dependencies on other
+workspaces from registry instead of using the local version. 
+
+It should still use the rootDir's lockfile since adding lockfiles for each workspace will cause conflicts
+across workspaces. Since other workspaces will not be in the lockfile, it should determine their versions as if it was 
+installing without a lockfile (by taking the newest version that matches their semver).
+
+#### Changes to current (non-isolated) behavior
+If a dependency was previously installed in a workspaceDir (from an isolated installation) and will be hoisted, delete it from the workspaceDir. This should not be a breaking change since the end result would be the same as if the isolated flag did not exist or if you were installing from scratch.
+
+### upgrade
+#### With isolated flag
+Upgrade all dependencies of the workspace in the root yarn.lock, respecting semvers of the current workspace and also
+every other workspaces. Install all dependencies for the current workspace in its node_modules, not hoisting anything
+to the root. Take the latest matching versions from the registry for all dependencies on other workspaces.
+
+#### Changes to current behavior. 
+Deleting modules installed at an individual workspaceDir if they are going to be hoisted (similar to changes for `install`).
+
+### upgrade-interactive
+#### With isolated flag
+Interactive version of an isolated upgrade. Bring up the interactive prompt, only showing the workspace's dependencies.
+Record upgrades in the root yarn.lock. Install all dependencies for the current workspace in its node_modules, not hoisting anything
+to the root. Take the latest matching versions from the registry for all dependencies on other workspaces.
+
+#### Changes to current behavior. 
+None except the changes to `upgrade`. Still just an interactive version of `upgrade`.
+
+### outdated
+#### With isolated flag
+Should show the list of packages that upgrade-interactive would show with isolated flag.
+
+#### Changes to current behavior. 
+None.
+
+### check
+#### With isolated flag
+Verify the workspace's package.json against the root yarn.lock file. Use the workspace node_modules (not hoisted) when looking
+at actual package contents.
+
+#### Changes to current behavior. 
+None.
+
+### list
+#### With isolated flag
+List only dependencies needed by the current workspace.
+
+#### Changes to current behavior. 
+None.
+
+# How We Teach This
+
+*What names and terminology work best for these concepts and why?*
+No new terminology is needed. The `--isolated` flag is just a new flag that certain commands recognize. The documentation should be able to give a good explanation of the behavior.
+
+*How is this idea best presented?*
+As a way of enabling isolated development in a single workspace without removing the ability to manage the whole
+workspace from a subfolder. 
+
+*How should this feature be introduced and taught to existing Yarn users?*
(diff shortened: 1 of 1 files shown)


accepted/0000-isolated-workspaces.md
```diff
@@ -0,0 +1,177 @@
+- Start Date: 2018-02-12
+- RFC PR: (leave this empty)
+- Yarn Issue: (leave this empty)
+
+# Summary
+
+Running a workspace in isolation should be as easy as running all workspaces together. Running a yarn command in
+a workspace should behave the same whether it is part of a larger project or whether it is its own separate repo. There
+should be a new set of commands, `yarn workspaces <command>` to distinguish between running commands that affect
+the whole workspace from commands that affect a single workspace.
```
bestander: nit: whole workspace => all workspaces

accepted/0000-isolated-workspaces.md
```diff
@@ -0,0 +1,177 @@
+- Start Date: 2018-02-12
+- RFC PR: (leave this empty)
+- Yarn Issue: (leave this empty)
+
+# Summary
+
+Running a workspace in isolation should be as easy as running all workspaces together. Running a yarn command in
```
bestander: nit: "Running a workspace" is a bit misleading, word missed?

accepted/0000-isolated-workspaces.md
```diff
@@ -0,0 +1,177 @@
...
+If you could work on a single workspace in isolation as easily as if it were its own repo, the transition to a monorepo
+from separate repos would be seamless for people who focus on a single workspace/repo at a time, and one of the
+biggest reasons not to use workspaces would be removed. 
+
+# Detailed design
+
+Terminology used for this doc:
+- rootDir: the root directory of a workspaces project. the one with the package.json with workspaces in it
+- workspaceDir: Directory for an individual workspace/package
+- regular (isolated) mode: `yarn foo`
```
bestander: I think isolated is a better term than regular

bestander:
@bdwain thanks for your RFC!

I was thinking of something way simpler.

Can we have a flag like `--isolated` that would just install/add/upgrade/ node_modules ignoring that there is a  workspace root and other workspaces as well?
Good point on reusing yarn.lock from the root though.

accepted/0000-isolated-workspaces.md
bdwain: agreed

accepted/0000-isolated-workspaces.md
bdwain: oops yea

accepted/0000-isolated-workspaces.md
bdwain: Maybe it could say 

Working on a workspace in isolation should be as easy as working on all workspaces together.

bdwain:
@bestander yea a flag on the existing commands also works. I kind of mentioned that in the unresolved questions section. I'm not too attached to having it be a separate command versus a flag. 

If possible though, I would like to retain the ability to run install in a specific workspace without a flag. The reason I think this would be useful is that for a team who is migrating away from a multi-repo setup, people would be used to working on projects in isolation, which means having yarn install behave in an isolated way. For a lot of these projects, it can be the case that the average person working on the project works on a single workspace at a time, and it would be annoying for them to always have to remember to type `--isolated` whenever running install/upgrade. By allowing them to run the same commands as before, it would make transitioning to a monorepo much simpler, and also just be easier in the long term if you work on isolated workspaces a lot.

I think the simplest way to achieve that would be what I described, where the commands are aware if they are run in a single workspace or at the root of the project. As a slightly more complicated alternative though, there could be 2 flags, `--isolated` and `--workspaces` and a default set for the project as to which mode you run in when no flag is set. The default can be in package.json or yarnrc or something, or maybe just a user-specific config value that is not expected to be in source control?

bdwain:
updated with your suggested changes

bdwain:
updated with the wording changes.

bestander:
BTW, did you know that you can set a flag via .yarnrc for any command?
https://yarnpkg.com/en/docs/yarnrc#toc-cli-arguments

The team that migrates to monorepo can just set
```
--install.isolated true
```
In their project's .yarnrc and it will do the trick.

accepted/0000-isolated-workspaces.md
```diff
@@ -0,0 +1,177 @@
...
+from separate repos would be seamless for people who focus on a single workspace/repo at a time, and one of the
+biggest reasons not to use workspaces would be removed. 
+
+# Detailed design
+
+Terminology used for this doc:
+- rootDir: the root directory of a workspaces project. the one with the package.json with workspaces in it
+- workspaceDir: Directory for an individual workspace/package
+- isolated mode: `yarn foo`
+- workspaces mode `yarn workspaces foo` (could also be `yarn foo --workspaces` or `yarn foo -W` like yarn add)
```
bestander: So this is the default mode right now and you propose isolated mode to become default?

I would rather avoid a breaking change and reverse it.
Via .yarnrc you can set up a subfolder to work in isolated mode and if a user cd to that folder Yarn would work in isolated mode.
But if user cd to any other folder Yarn would work the way it works right now

bdwain: That works. I did not know about settings flags via yarnrc. I do think it's worth reconsidering the default mode at the next major version, but that can be a separate issue.

So if i used that, i could still go into the workspace and say `yarn install --isolate=false` to run explicitly in workspaces mode? That is a bit awkward. Kinda why i was thinking of having a --workspaces flag that is basically the opposite of --isolate.

bestander: I propose doing a non breaking progressive change first and shipping it.
Then before Yarn v2.0 we can ask people to weigh in if introducing the reverse would be better for them.

bdwain: fair enough! I'll make those changes to the doc

bdwain:
@bestander i changed the proposal to have an --isolated flag

bestander:
Thanks, @bdwain, I think we achieved a nice balance here.
I am excited to see the PR with this feature, certainly useful for many cases.

bdwain:
Sweet! I can try to work on a PR to implement this. Thanks for the feedback!

bdwain:
@bestander I've gotten `yarn install --isolated` working locally. there's more work to do (other commands and tests and everything), but i just wanted to make sure what I've got so far looks to a good general approach since it's a non-trivial change and I don't have much experience with the yarn codebase. I don't want to get too far in only to find out the approach is all bad.

https://github.com/yarnpkg/yarn/compare/master...bdwain:isolated

Do you think you could check out that branch (it's small) and let me know if my general approach looks alright? Or if you're busy if you could point me to someone else who might be able to look? 

Thanks!

choffmeister:
I am looking forward for this to be supported by yarn natively. Until then I wrote a little package that provides a way to isolate a single yarn workspace:

https://github.com/choffmeister/yarn-workspace-isolator

So far it worked pretty well for me (I used it to build individual docker images from my monorepo), but there are for sure some edgecases around the corner.

Maybe this can help others, too, until this RFC is implemented.

bestander:
@bdwain, looks good, can't wait to see the PR!

marcinczenko:
> Can we have a flag like --isolated that would just install/add/upgrade/ node_modules ignoring that there is a workspace root and other workspaces as well?

Hmm...maybe you already cover that, but I do not see it working for the following case.

Say I am having a monorepo:

```
packages
    package-1     
    package-2
node_modules
    ...hoisted modules
package.json
```

Imagine this monorepo is to be open-sourced, but at the same time you start working on a `client-app`. This `client-app` is not supposed to land in the monorepo, but it has its own repository. When working on it, you will simultaneously work on packages `package-1` and `package-2`, which are dependencies for the `client-app`, so you would like to take advantage of the monorepo ecosystem to speed up the development.

Thus, imagine you clone the `client-app` repo to the `packages` folder. The `.gitignore` of the mono repo ignores everything except `package-1` and `package-2`, so anything you do with the `client-app` will only be visible to its own repository.

Now I want to be able to run `yarn install`. I would like to get this:

```
packages
    client-app
        node_module
            package-1 => symlink to package-1
            package-2 => symlink to package-2
            ...all other dependencies are here and not hoisted
    package-1        
    package-2
node_modules
    ...hoisted modules
package.json
```

So in other words I would like that yarn install behaves as usually with respect to symlinking, but does not hoist any other modules that are listed as dependencies of `client-app`. In principle, it would be possible to hoist other dependencies from the `client-app`, but in this specific case we do not want it because it would leave trace on the root `yarn.lock`. The `client-app` is here just for the convenience of the development workflow, but it will be removed at some point. We also cannot accept situation in which we make it part of the monorepo repository just temporarily - `client-app` has a license that does not allow it to be published in a public repository (which is ok for the monorepo).

I would imagine that in the root `package.json` one could say:

```
hoisting: {
    excludeWorkspaces: [
        'client-app'
    ],
    /// other hoisting related options
}
```

Does it make sense, or maybe you think this should be approached in some other, better way?

bdwain:
my thought is that monorepos aren't really designed to have arbitrary consumer packages added temporarily. if the packages are closely related enough that someone would develop them locally against eachother, it probably makes sense for them to be in the same monorepo. If it's just something like a client-app developer is making a PR to the monorepo and wants to test their changes locally (but does not normally work on it), they can always use yarn link to try it out.

marcinczenko:
@bdwain ok, I should not mention "temporarily" - the key point in my initial comment is that the `client-app` has more restricting license than the remaining part of the monorepo and cannot be exposed in the same archive. It may happen that they will stay together for a very long time. This, using `yarn link` is not a solution.

bdwain:
I'm a bit confused. Is the monorepo aware of client-app? Or does it only exist inside the monorepo when working locally? Is this needed to simplify development?

Git submodules/subtrees/[subrepos](https://github.com/ingydotnet/git-subrepo) may be a useful tool for something like that either way. I've heard mixed things about them, but they might be useful for adding arbitrary other repos into the monorepo without committing the actual code.

marcinczenko:
Yes, it may be a bit confusing, but this is actually very common scenario in our environment (mixed Open Source and not open source). So monorepo is both aware and not aware of the `client-app`. It is aware in the way that it should respond to all automatic symlinking but at the same time - git-wise - `client-app` should not leave any traces on the monorepo's  git. This means that submodules/subtrees   /subrepo will not work I am afraid because the leave trace in the parent repo (correct if I am wrong wrt to subrepo as I did not really use that one). You mentioned before `yarn link` - so this would work, but it is quite inconvenient - we like to take advantage of yarn handling this stuff automatically.

marcinczenko:
I would not say that it is a total opposite of an isolated workspace (in some way it is), but the rest you got right :).

marcinczenko:
Yarn is introducing the `nohoist` option for workspaces: https://yarnpkg.com/blog/2018/02/15/nohoist/. Unfortunately, this will not let you ignore whole workspaces from being hoisted as described above.

bdwain:
@marcinczenko the proposal ended up changing a bit. Check out the focused workspace folder in the accepted folder. It’s a bit simpler now. I think nohoist should work well with it.

PR #91, opened 2018-03-01
Isolated Workspaces Updates

bdwain:
@bestander I have some small changes I realized were necessary after beginning work on this.

[rendered](https://github.com/bdwain/rfcs/blob/master/accepted/0000-isolated-workspaces.md)

### DIFF
--- a/accepted/0000-isolated-workspaces.md
+++ b/accepted/0000-isolated-workspaces.md
@@ -32,6 +32,7 @@ relevant yarn commands should be able to be run with an `--isolated` flag.
 
 Relevant commands (based on current functionality) include
 - install
+- add
 - upgrade
 - upgrade-interactive
 - outdated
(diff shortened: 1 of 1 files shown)


accepted/0000-isolated-workspaces.md
```diff
@@ -46,37 +47,30 @@ If you run a command with the isolated flag and are in a workspaceDir, it will r
...
 ### install
-#### With isolated flag
 Install only that package's dependencies, and do not hoist them to the rootDir. Install dependencies on other
 workspaces from registry instead of using the local version. 
 
 It should still use the rootDir's lockfile since adding lockfiles for each workspace will cause conflicts
 across workspaces. Since other workspaces will not be in the lockfile, it should determine their versions as if it was 
 installing without a lockfile (by taking the newest version that matches their semver).
 
-#### Changes to current (non-isolated) behavior
```
bdwain: deleting the dependency on install seems to happen automatically already, so it was unnecessary. With that, no existing commands needed changing.

bdwain: deleting the dependency on install seems to happen automatically already, so it was unnecessary. With that, no existing commands needed changing.

bestander:
Looks good, thanks @bdwain!

PR #92, opened 2018-03-07
isolated workspaces updates

bdwain:
@bestander I've been working on isolated workspaces for a bit and I've got some more thoughts now that I have a little bit of a better idea of what yarn's codebase looks like and what the RFC's implementation would need to be like. I've also got a potentially much simpler proposal that may be much better overall (it's at the bottom if you want to skip to it). I made this PR to start a new discussion since I didn't think it made sense to reuse the old PR and there's no issues in this repo, but it's just an empty change for now. And here's my current branch if you're curious where I left off https://github.com/yarnpkg/yarn/compare/master...bdwain:isolated

## Original Idea Updates
In hindsight, I think the original isolated workspaces proposal may have been a little ambitious and difficult (if not impossible) with the way workspaces are currently implemented. Specifically, working with the lockfile in an isolated fashion seems very difficult. Getting a basic `yarn install --isolated` working that ignored the lockfile wasn't so bad, but most of the other commands (which don't make sense if you ignore the lockfile) look like they will be much more complex. It's probably possible, but I've been spending a lot of time trying to understand the package resolution/fetching/linking logic and while I don't have a great grasp of it yet, I'm pretty sure there would be significant changes needed to allow sensible upgrades to the lockfile to occur that respected other workspaces. 

I've been considering how a scaled down approach (at least at first) might work, and maybe we could try just allowing --isolated on install, but I'm worried about how my current implementation kind of abandons reproducible builds by ignoring the lockfile completely.

##  New Idea
I had a much simpler idea tonight that may make all of the above unnecessary. There's not really a need to install everything locally inside a workspace since you already have it at the root node modules directory from a non-isolated install. So instead of trying to make every command work in isolation, what if we just added one new command, `yarn isolate`. It's easiest to explain it with an example.

Let's say i have a monorepo with packages A,  B and C, where A depends on B and C (and any other arbitrary dependencies). If i try to run A in isolation, `yarn isolate` will just shallowly install B and C in `A/node_modules`. By shallowly install, I mean install them but not install any of their transitive dependencies. A regular install would already guarantee that those transitive dependencies would be at `root/node_modules`, so the only extra things isolation would really need would be sibling workspaces (and avoiding building those is basically the entire point of isolation anyway).

The benefits of this approach are:
- No existing commands need to change or have new flags added to them.
- Isolate is one simple new command to learn, and what it does should have no effect on the lockfile,
 which should help keep it from getting too complex
- It's easier to switch from isolation to non-isolation and back since all that needs to happen is you add or delete your workspace dependencies from the package's node modules folder (i.e. add or delete `A/node_modules/B and C`)

Potential downsides are
- You have to rerun isolate every time you run install/upgrade/add (since they would delete your isolated copies of workspaces)
  - This is easily solvable with command aliases/yarnrc settings/package.json scripts
- There is some potential complexity if A and B share a dependency (let's say lodash) but have different versions of it. You would need to pull down B's version of lodash and add it to `A/node_modules/B/node_modules/lodash` to guarantee correctness (since it would not be at the root node_modules)
  - This is still much less complexity than the original proposal
- installing in isolation from scratch is a bit slower now because you still need to do the full install
  - This should not be terribly slow since workspaces optimizes the install already by hoisting shared stuff and only installing it once

## Conclusion
Sorry for all of the flip-flopping. I didn't really understand the complexity of the original proposal until I got into the code. 

If you think the new proposal makes more sense, I can update the RFC and we can do whatever needs to be done to get started working on it. If not, I'd appreciate your thoughts on how to move forward with the original proposal and how we can address the issues I mentioned.

Thanks!


bestander:
At a high level looks smart.
Give me a couple of days before I can look at this thoroughly, if the
implementation is easy then go ahead and build a prototype.
Thanks for putting so much thinking into it

bdwain:
Cool yea i'll play around with it until then and if i can get something working I'll post it here.

bdwain:
fyi i've been trying to get it working by taking advantage of the nohoist logic (at least as a first attempt), though I haven't been successful yet. I'm still trying to wrap my head around all of the stages of an installation and how everything works.

bdwain:
@bestander I have a prototype that doesn't fully work yet but seems to get the correct files in the correct place. I'd love to get your thoughts on it. 

https://github.com/yarnpkg/yarn/compare/master...bdwain:isolated-v3#diff-15cb3e2b8427a3bddb8c3d74e1867c10

I think this general approach should work. It installs only the needed workspaces shallowly, and installs any dependencies in their node modules directories that can not use the hoisted version. I also had it do a full install so that it can be run instead of yarn install if desired, rather than having to run both. 

The way it works in general is it takes the hoisted tree that was calculated and figures out which workspaces depend on the one you want to isolate and which dependencies they need from the hoisted tree. It then just adds entries to the end of the tree that aren't hoisted.

I haven't touched the package install scripts part yet, so that will add some more work. There also seems to be some timing issue because it fails if i don't run it in debug mode. But in general I think this is working well (definitely much better than the other approaches I had)

bdwain:
@bestander I updated the RFC in this PR to reflect the new design I described ([here's the rendered view](https://github.com/bdwain/rfcs/blob/master/accepted/0000-isolated-workspaces.md)). I think my POC confirmed my guess that this is the right approach. It was much simpler overall than the original proposal and should be a better user experience also because it's so much faster.

I also changed it to a flag on the install command rather than a whole new command (my branch is still a command though). It's pretty similar to install, so it seemed to make more sense as a flag rather than a whole new command. I'm not too attached either way though if you want to change it. I explained the choice in the RFC also.

bdwain:
The last edge case I can think of is kind of a weird one. The desired behavior is not 100% obvious.

Normally, if workspace A depends on workspace B, and B is currently version 1.2.3, isolated mode would pull down B@1.2.3 from npm as long as its semver for B was satisfied by 1.2.3. 

Where this gets weird is if if B's package.json has unpublished changes but is still 1.2.3. Then the local B@1.2.3 has a different package.json from the remote one. 

- If we used the local package.json to determine all dependencies, we could be missing a dependency if the remote version of package.json still depends on one that was removed. 
- If we use the remote version to determine all dependencies, we would need to reinstall if we wanted to use B locally after running A in isolated mode. 
- If we want to handle both independently, it would require significant changes to the code because package name and version would no longer be a unique identifier, and there's a few places where this is assumed to be unique. 

I'm leaning towards option 2, at least initially. It's not ideal, but this seems like a rare enough scenario that it may not be worth the extra complications that trying to do option 3 will require.

bestander:
@bdwain, sorry for a delay, I was in an intense "war room" project that took all my time and I had to not pay attention to emails.
I am back now :)

bdwain:
Sweet! 

So this pr is up to date with the latest changes to the rfc, and the branch I linked above mostly works (it’s still a command because that’s easier for testing. I’ll switch to a flag later). I also did what I talked about in my previous comment. Haven’t done any tests yet though. I’ll do those later. 

It doesn’t do bin links correctly yet for the new packages And it doesn’t run post install scripts yet on the new packages, but other than that it works. There’s a few slightly hacky things but overall i think the design generally works. 

Please let me know your thoughts on the approach when you get a chance. Thanks!

bestander:
Hey @bdwain, I like that you considered making a simpler approach, simpler is always better.
What do you think of word `focus`?
Like `yarn focus` or `yarn workspace focus` would do the thing you described above.
Isolate seems a bit misleading to me now

bestander:
So if I understand the above correctly, this is the idea:

1. We run `yarn install` as usual for the whole monorepo 
2. but then we want to work on one of the workspaces (A) "isolated" as we don't want to deal with "dev mode" of other workspaces, 
3. so we go to the workspace A folder run `yarn isolate` there and it only installs all other dependent workspaces (B and C) form registry into node_modules.
4. If B and C have packages that were not hoisted to the root because of conflicts with other workspaces' dependencies they will be added to the A's node_modules too

Seems legit and should be doable.
Probably worth thinking more what happens when you run `yarn add/install/remove` in root after isolating.

Sorry again for being on and off, I am helping with Yarn on my spare time right now.

bdwain:
Thanks for the feedback!

For the name, i think isolate and focus both work, so if you prefer focus we can go with that. I would like to keep it as an option on install if possible because the functionality is so similar, it seems odd to make it a whole new command. Either way, the implementation should be similar. 

Your understanding of the behavior is correct. Basically, install sibling workspace dependencies locally and then handle all version conflicts/edge cases in a way that to ensures correct and consistent behavior. Number 4 in your list was one of those edge cases, but there are others. 

As for other commands, I think their behaviors should remain unchanged (which means they would delete isolated modules if run) to avoid confusion. But if we wanted, we could add —focus options to them too
So that they caused the isolated modules to be installed or stay installed. Using an option rather than a command would be helpful for that as well.

bdwain:
Sounds good! This is good to merge except the name is still isolate

bdwain:
Ok updated to be named focus!

bestander:
Thanks, @bdwain

PR #95, opened 2018-08-01
Proposal to improve yarn why

VincentBailly:


accepted/0000-better-why.md
```diff
@@ -0,0 +1,142 @@
...
+It would be much more useful to show the dependency chains leading to this package.
+
+# Motivation
+
+Knowing which direct dependencies cause a given package to be installed is
+useful in several cases.
+
+## Really understand why a dependency got installed
+
+The current version of `yarn why` does not really tells why a dependency is installed,
```
bestander: type: tell

bestander:
I think it is a very reasonable feature.
Thanks for sending the RFC, LGTM

PR #99, opened 2018-08-21
Rewrite the `registry-url-in-lock-file` RFC

Gudahtt:
The `registry-url-in-lock-file` RFC was a proposal for allowing users to more easily switch between registries. The current problem with switching registries is that Yarn tries to use the cached tarball URL in the `resolved` field of the lockfile instead of the current registry when installing.

The original proposal was to remove the registry from the lockfile. This would force Yarn to use the current registry. Later in the discussion it was also proposed to follow the behavior of `npm@5` and just ignore the `resolved` field.

Both of these solutions seemed to have the support of everyone involved in the discussion, but the concern of this being a breaking change was raised. So instead, a third proposal for an additional configuration field was what ended up being merged.

Circumstances have changed, and a major version bump is likely in the near future. Given that we can now make a breaking change, I am proposing that we rewrite this RFC to more closely match its original intention. This rewrite proposes that we solve the problem by adopting the same behavior as `npm@5`.

I have also updated the title of the RFC to better describe the proposal.


accepted/0000-registry-url-in-lock-file.md
```diff
@@ -1,102 +0,0 @@
...
-- Yarn Issue:
-
-# Summary
-
-The yarn.lock file includes information about the registry that it was generated against. It would
-be nice if there was no registry information associated with the lock file
-
-# Motivation
-
-We currently have two internal registries/mirrors. We generate the lock file against one we have in
```
Gudahtt: I did make an attempt at including this use case, but I didn't understand it well enough to explain it.

accepted/0000-switching-registries.md
```diff
@@ -0,0 +1,63 @@
...
+
+The `resolved` field will not be used to determine which registry is used. Changes to to the `resolved` field, such as removing the registry, are outside the scope of this RFC.
+
+### How do scoped registries work?
+
+Yarn supports configuring a registry for a specific scope (e.g. `yarn config set '@foo:registry' 'https://registry.foo.com'`). If a scoped registry has been configured, then this registry shall be used for **all** packages under that scope. Using different registries within a single scope is not supported.
+
+### Can non-scoped packages use alternate registries?
+
+No, non-scoped packages would all use the same registry. This restriction simplifies the configuration, and keeps us in line with `npm`.
```
arcanis: Well, they can use alternate registries, but on a "all or nothing basis". The technical answer is "yes" but your explanation is good 😛

accepted/0000-switching-registries.md
```diff
@@ -0,0 +1,63 @@
...
+# Detailed design
+
+Yarn should adopt the behavior `npm` introduced with `npm@5`, which is to always use the registry configured by the current user. This change was described in [this blog post](http://blog.npmjs.org/post/161081169345/v500):
+>If you generated your package lock against registry A, and you switch to registry B, npm will now try to install the packages from registry B, instead of A. If you want to use different registries for different packages, use scope-specific registries (npm config set @myscope:registry=https://myownregist.ry/packages/). Different registries for different unscoped packages are not supported anymore.
+
+Yarn already supports switching to a different registry and scoped registries. The change would be to use them in all cases rather than just for new packages.
+
+### What about the `resolved` field in the lockfile?
+
+The `resolved` field will not be used to determine which registry is used. Changes to to the `resolved` field, such as removing the registry, are outside the scope of this RFC.
```
arcanis: Not that much - do you see a drawback removing it from the lockfile? Leaving it there seems confusing, since the lockfile would then contain "useless" data.

accepted/0000-registry-url-in-lock-file.md
arcanis: I think this use case is largely solved by one of those two options:

- Users could simply have a CI build step where they copy `.yarnrc.ci` into `.yarnrc`, or
- Users could call Yarn with the `--use-yarnrc .yarnrc.ci` flag

This would let them define the registries they want to use on CI without too much hassle.

accepted/0000-switching-registries.md
Gudahtt: Ah.. right 😛  I shall edit this to make that more clear.

bestander:
I love the proposal, thank you @Gudahtt  for making it backwards compatible.
@arcanis approved it. I'll merge then

bestander:
@Gudahtt PR coming? ;)

Gudahtt:
Sure! I'll start work on it this weekend.

accepted/0000-switching-registries.md
hbetts: @Gudahtt thank you for driving this work forward.

This RFC will help our organization work around the consequences of hard-coded registry URLs in the lock file.

PR #101, opened 2018-09-13
Yarn Plug'n'Play: Getting rid of node_modules

arcanis:
> **Note:** This PR's discussions should be focused on the high-level design. [A separate PR] (https://github.com/yarnpkg/yarn/issues/6382) has been opened on the code repository to discuss the implementation details.

> A pdf has been generated and made available [here](https://github.com/yarnpkg/rfcs/files/2378943/Plugnplay.pdf) for easier reading. Note that this PR remains the reference location for the most up-to-date information regarding this proposal.

Hi folks!

We propose in this RFC a new alternative and entirely optional way to resolve dependencies installed on the disk, in order to solve issues caused by the incomplete knowledge Node has regarding the dependency tree. We also detail the actual implementation we went with, describing the rational behind the design choice we made.

I'll keep it short here since there's much to discuss in the document itself, but here are some highlights:

* Installs ran using Plug'n'Play are up to 70% faster than regular ones ([sample app](https://github.com/yarnpkg/pnp-sample-app))
* Starting from this PR, Yarn will now be on the path to make yarn install a no-op on CI
* Yarn will now be able to tell you precisely when you forgot to list packages in your dependencies
* Your applications will boot faster through a hybrid approach of static resolutions

This is but a high-level description of some of the benefits unlocked by Plug'n'Play, I encourage you to give a look at the document for more information about the specific design choices - and in case anything is missing, please ask and I'll do my best to explain them more in depth!

I should mention that we've been using in production inside Facebook for about two weeks now, and didn't get issues since then. Now that it passed the trial by fire we felt confident enough that this solution was the right solution, and share it openly so that we can all iterate on it.

Working on this project has been super exciting for me, and I can't wait to see the new possibilities that it will unlock! Especially from a tooling perspective, the benefits of having a unified indirection allowing package managers to dictate the way the dependency are loaded unlocks new incredible patterns and make it easier and safer for tools to integrate with it.

Paging some community members that have been made aware of the project during its development and helped us in various ways, either through actual contributions (kudos to @imsnif for implementing `yarn unplug`!) or by their feedback:

- Yarn maintainers: @imsnif / @rally25rs / @kaylie-alexa / @BYK 
- Tool authors: @TheLarkInn / @sokra / @devongovett / @Rich-Harris / @not-an-aardvark
- Package managers: @iarna / @zkochan
- Node foundation: @boneskull / @MylesBorins

### DIFF
--- a/accepted/0000-plug-an-play.md
+++ b/accepted/0000-plug-an-play.md
@@ -149,7 +149,7 @@ In this situation, even though it is found multiple times inside the dependency 
 
 Since Plug'n'Play flattens the dependency tree while still preserving the links between the nodes, the paths Node will get will be the same for any `package-c@1.0.0` inside the dependency tree, causing the package to be instantiated a single time.
 
-### A. users cannot require dependencies that aren't listed in their dependencies
+### E. users cannot require dependencies that aren't listed in their dependencies
 
 A common problem was that it was extremely easy for a library author to start relying on a package and forget listing it inside the dependencies. Because these broken dependencies were being pulled by dev dependencies before being hoisted to the top level, they often happened to work fine in development environments and break in production.
 
@@ -216,7 +216,7 @@ Some packages try to require packages they don't directly depend on for legit re
 
 In order to solve this, Plug'n'Play details a special case when a package makes a require call to a package it doesn't own but that the top-level has listed as one of its dependencies. In such a case, the require call will succeed, and the path that will be returned will be the exact same one as the one that would be obtained if the top-level package was making the call.
 
-### e. edit-in-place workflows need different tools
+### E. edit-in-place workflows need different tools
 
 A quite common debug pattern is to manually edit the `node_modules` files in order to alter the behavior of the program and debug it more easily (by adding `console.log` statements, for example). Since the `node_modules` folders don't exist anymore, this isn't directly possible anymore.
 
--- a/accepted/0000-plug-an-play.md
+++ b/accepted/0000-plug-an-play.md
@@ -1,0 +1,302 @@
+- Start Date: 13 Sep 18
+- RFC PR: https://github.com/yarnpkg/rfcs/pull/101
+- Yarn Issue: https://github.com/yarnpkg/yarn/issues/6382
+- Champion: Maël Nison (@arcanis - [Twitter](https://twitter.com/arcanis))
+
+# Plug'n'Play Whitepaper
+
+## 1. Summary
+
+We propose in this RFC a new alternative and entirely optional way to resolve dependencies installed on the disk, in order to solve issues caused by the incomplete knowledge Node has regarding the dependency tree. We also detail the actual implementation we went with, describing the rational behind the design choice we made.
+
+I'll keep it short in this summary since the document is already large, but here are some highlights:
+
+* Installs ran using Plug'n'Play are up to 70% faster than regular ones ([sample app](https://github.com/yarnpkg/pnp-sample-app))
+* Starting from this PR, Yarn will now be on the path to make yarn install a no-op on CI
+* Yarn will now be able to tell you precisely when you forgot to list packages in your dependencies
+* Your applications will boot faster through a hybrid approach of static resolutions
+
+This is but a high-level description of some of the benefits unlocked by Plug'n'Play, I encourage you to give a look at the document for more information about the specific design choices - and in case anything is missing, please ask and I'll do my best to explain them more in depth!
+
+## 2. Motivation
+
+When the first Javascript package managers appeared, they had to use the tools available to them. As a consequence, they were implemented on top of the Node resolution algorithm, copying the packages in such a layout that Node would be able to find them without external tool. This allowed the ecosystem to thrive, which interestingly revealed some of the scalability flaws in this approach:
+
+* The Node resolution algorithm isn't aware of what packages are, and as a result doesn't know anything about dependencies either. When a module makes a require call, Node simply traverses the filesystem until it finds something that matches the request, and uses it.
+* Installations take time, and the time required to copy files from the package managers' caches to the `node_modules` folders is one of the main bottleneck as it is heavily I/O bound, which cannot be easily optimized. Copy-on-Write filesystems can alleviate this issue to an extent but come with their own set of drawbacks, such as a lack of global support.
+
+We at Facebook suffered from these issues, and decided to find a way to solve them cleanly while staying compatible with the current ecosystem. The solution we found has been put to the test inside our infrastructure for a few weeks now, and we now feel confident enough about its merits that we want to share with the community at large, and continue iterating on it in the open.
+
+## 3. Detailed Design
+
+In its current state, running `yarn install` does the following under the hood - assuming a cold setup:
+
+1. Dependency ranges are resolved into pinned versions
+2. Packages for each version are fetched from their sources and stored in the offline mirror
+3. The offline mirror is unpacked into the cache
+4. The cache is copied into the `node_modules` folders
+
+Our solution aims to trim the fourth step from the equation. Instead of copying each and every file from the cache to the `node_modules` folder (which can take a lot of time, amplified by the sheer number of files), we will instead generate a single file that will contain static resolutions tables that list:
+
+* What packages are available in the dependency tree
+* How they are linked together
+* Where they are located on the disk
+
+A special resolver is then able to leverage the knowledge extracted from those tables to guide Node and help it figure out the location where each package has been installed (in our case, the Yarn cache). Since all packages from the dependency tree can be statically found inside the tables, the whole filesystem traversal required by the `node_modules` resolution can be skipped - bringing a free performance win at runtime by decreasing the amount of filesystem I/O needed to boot Node applications.
+
+> **Why a special resolver rather than symlinks/hardlinks?**
+>
+> While quite useful, we believe both symlinks and hardlinks are solving the wrong problem. Symlinks require tooling support and have ambiguous semantics (Node has no way to know whether a symlink has been created through yarn link, through a workspace, or through a special installation). Hardlinks have surprising behaviors (they unexpectedly corrupt your cache if you change them), don't play well with cross-volumes installs, and also require tooling support (for example when trying to persist the `node_modules` folders). Both of them require heavy I/O at install time, don't decrease the I/O at runtime, and more generally try to workaround the Node resolution rather than change it.
+
+In the end, while some package managers had some success with these strategies (shoutout to pnpm in particular which was one of the first package managers experimenting to solve those problems!), we envision an alternative approach that's working for us at large scale and that we hope will work for others as well.
+
+### generated api
+
+A point needs to be made regarding what Plug'n'Play is. While the previous section described the way Yarn would now generate “static resolution tables”, those tables are not what make Plug'n'Play what it is - instead, they are an implementation detail amongst others. Instead, Plug'n'Play is first and foremost an API meant to abstract the resolution process and provide strict behavioral guarantees. In consequence, the resolution tables are useless without the resolver itself, since it's the resolver that is standardized, not the tables.
+
+For this reason, we decided to generate a single Javascript file (called `.pnp.js`, for “Plug'n'Play”) that would contain both the static resolution tables and the resolver itself. This has multiple advantages:
+
+* The first one is encapsulation. By keeping the static resolution tables private to the resolver, we also prevent third-party scripts from relying on them, thus making it possible for us to change their underlying implementation as we see fit. A real-life case study lies within the `findPackageLocator` function. Its current implementation is somewhat naive and would be much improved through the use of a [trie](https://en.wikipedia.org/wiki/Trie). Since the static resolution tables are hidden, we can easily replace their format to match this new data structure, which wouldn't necessarily be possible if we had publicized the underlying data structures.
+* Project dependencies are traditionally versioned through the `dependencies` field, and installed using `yarn install`. But assuming that the resolver would be kept within Yarn, how would we make sure that the generated file is always compatible with the current Yarn version? In fact, how would we make sure that Yarn itself was available? As detailed in a later section, one of the future improvements we've planned is to make Yarn entirely optional on CI. In this context, where would the “runtime” required for such a resolver be located? The easiest answer is to keep the static resolution tables tied to the resolver itself, removing the risk of unexpected incompatibilities.
+
+As an executable Javascript file, the generated `.pnp.js` file presents the following features:
+
+* It doesn't have any dependency other than the built-in Node modules - it doesn't even depend on Yarn itself.
+* It exposes an API that can be used to programmatically query resolutions without having to parse the file or to load the full resolver. As mentioned, the implementation itself is entirely free as long as both the interface specified in Annex B and the contract listed in Annex C are fulfilled. This should leave enough room for experimentation while providing a consistent API that matches the current expectations of most packages.
+* It's an executable script that can act as a resolution daemon that other processes can communicate with through standard input / standard output using the protocol listed in Annex D (based on JSON). This makes it suitable for integration with third-party tools not written in Javascript - this allowed us for example [to introduce the `module.resolver` option into Flow](https://github.com/facebook/flow/commit/7b6738bdba7b6a4a5844c079d9dd1ddcf815effb). Flow is written in OCaml and cannot use the Javascript API, but thanks to this small bridge, it was only a matter of parsing a few lines of JSON.
+* If loaded as a preloaded module (`node -r ./.pnp.js`), it will inject itself into the Node environment and will transparently cause the `require` function to load files directly from the right location. The current implementation overrides `Module._load`, but Node 10 recently released a new API that we plan to use to register into the resolver.
+* While not guaranteed strictly speaking (should it?), the `.pnp.js` file implemented by Yarn is stable through the use of relative paths. It means that the file can be moved from a computer to another and will still work as expected (provided that the cache folder on the new environment is both hot and located in the same path relative to the `.pnp.js` file). Through smart uses of the `.yarnrc` argument options, it becomes possible to store both the cache and the `.pnp.js` files together, allowing to skip the installs altogether and making Yarn optional.
+
+### workspaces & links
+
+Yarn supports adding persistent symlinks to your projects through two means: the first one, which we recommend, is to use the workspaces feature in order to create automatic links between your packages. The second one, which is a bit older, is to use the `link:` protocol and force Yarn to create a symlink to a location, regardless what's located there.
+
+Both of those will continue working with Plug'n'Play, but will be slightly repurposed: when operating under a Plug'n'Play environment, the links between the packages will be kept inside the `.pnp.js` file, and only there. It means that we won't be creating actual symlinks anymore - we don't need them, since their main goal was to hook into the require process! In case a user needs to access something “through the symlink”, they just have to use `require.resolve`, which will query the Plug'n'Play resolution and return the right path.
+
+A third way of adding symlinks also exists in traditional installs: `yarn link`. While it would be possible to implement something similar, we decided not to rush it since it comes with additional issues. In case you need to use yarn link:
+
+* If needed for debugging a project, we recommend to use `yarn unplug`, which will copy a package from your cache into the `.pnp/unplugged` folder. This “unplugged” copy is entirely free for you to alter as you see fit. Once you're done, just run `yarn unplug --purge-all` and the modifications you've made will be forgotten. Note that **this feature is not meant to be used as a permanent trick.**
+* If you were using `yarn link` as part of your actual install process, we recommend you to either use the `link:` protocol (but be aware of the issues caused by split dependency trees, cf Section 5.B) or, much better, to port your project to use workspaces (which don't suffer from the issues described in Section 5.B since Yarn is then aware of the whole dependency tree).
+
+### Virtual packages
+
+> **Packages instantiations**
+>
+> As a reminder, a module instance is the representation in memory of this module (it's usually the `module.exports` value exported by this value). Node caches each require call so that if a same module is required multiple times, it will only be instantiated once. The way it does this is by comparing the [realpath](https://nodejs.org/api/fs.html#fs_fs_realpath_path_options_callback) of the files. Unfortunately, this heuristic doesn't work when a same package is located in multiple different locations - which can happen when a package cannot be hoisted, for example. In this instance, Node will create one separate cache entry for each time the package has been duplicated, increasing the time needed for the application to start and messing with `instanceof` checks.
+
+
+Plug'n'Play guarantees that each combination of package name / version will only be instantiated once by Node, **except in one documented case**: if a package has any number of peer dependencies, Node will instantiate them exactly once for each time it is found in the dependency tree - and this regardless of whether the packages are strictly identical or not.
+
+So for example if you have `react` and `react-dom` (which has a peer dependency on `react`), a same version of `react` will be guaranteed to only ever be instantiated once by Node (because it doesn't have any peer dependencies), but `react-dom` will be instantiated exactly once for each package that depends on it.
+
+
+> **Why does it work this way?**
+>
+> Let's say you have `package-a` and `package-b`. Both of them depending on the same package `child`, which has a peer dependency on `peer`. Now, imagine that `package-a` depends on `peer@1` while `package-b` depends on `peer@2`. In this instance, the `child` package will have to be instantiated twice in order for us to satisfy the peer dependency contract for both `package-a` and `package-b`. We do this by computing a “unique identifier” (also called a virtual package) for each package instance. Now, how can we standardize the wording for this behavior?
+>
+> First solution would be to say something similar to: “a package with peer dependencies must have exactly one unique identifier per each set of inherited dependencies”. So in our example, since `child` has two sets of inherited dependencies, it would get two unique identifiers, would be instantiated twice, and everything would work. But unfortunately it's not so simple.
+>
+> Problems arise when you consider circular dependencies. Let's imagine a different scenario: `package` depends on `child-a` and `child-b`. The `child-a` package has a peer dependency on `child-b`, and `child-b` has a peer dependency on `child-a`. In this situation, per the wording described above, we would need to generate the unique identifier for `child-a` based on the set of its dependencies, which includes `child-b`. The problem is that the unique identifier for `child-b` has not been generated yet, so we cannot compute the unique identifier for `child-a`, and vice-versa! The loop cannot be broken.
+>
+> The solution to this issue is to say that the unique identifier for a package with peer dependencies is based on the unique identifier of its direct parent. Since a package unique identifier is always computed before its children, we cannot have a cyclic dependency.
+
+### Install config
+
+Since this proposal is still experimental we decided not to enable Plug'n'Play by default for the time being. As a result, a new key has been added to the `package.json`: `installConfig` (we selected this name to mirror the `publishConfig` settings that already existed). If `installConfig.pnp` contains a truthy value, Plug'n'Play will be used if the current version of Yarn supports it. Otherwise, the regular install will be used:
+
+```
+{
+  "installConfig": {
+    "pnp": true
+  }
+}
+```
+
+> **Why not store this configuration value within the `.yarnrc`?**
+>
+> While the `.yarnrc` files would have made a fine candidate, we believe that checking whether a project is Plug'n'Play compatible or not can be extremely interesting for various tools. Listing it into the `package.json` means that virtually any tool can quickly decide whether they want to take advantage of the guarantees provided by Plug'n'Play.
+
+## 4. Solved Issues
+
+### A. INSTALL SPEED NOW REACHES NEW ALL-TIME-HIGHS
+
+Probably the most obvious win is that by keeping the dependencies files stored into the cache, Yarn doesn't have to copy them around anymore. This allows the link step to be skipped almost entirely - the last remaining actions being done only because of how Yarn is currently architectured and will be removed later on (we're still creating a deep tree before flattening it in a later step, which is unnecessary).
+
+### B. Installs can now be efficiently cached even on ci
+
+Now that files don't have to be copied anymore, efficient caching becomes a breeze. The actual Yarn cache can be persisted into locations shared between all CI instances, making it possible to skip installs altogether provided both the cache and the `.pnp.js` file are made available.
+
+Where some projects were spending more than two minutes running (like is the case for react-native, for example), Plug'n'Play now allows them to spend this time actually running their tests, decreasing the load on the CIs and making for a better developer experience - we all hate waiting for the tests to end.
+
+### C. users working on multiple projects across a system won't pay increasing install costs
+
+A common occurence in the Javascript world is developers working on multiple disconnected projects sharing similar dependencies (for example all projects created through `create-react-app`). Due to how package managers currently work, the files used by those projects were typically copied from the cache into multiple `node_modules`, multiplying both the total size of the installs on the disk and the time wasted running installs.
+
+Now that the files are read directly from the cache, no matter your system, you'll only ever pay the cost of having multiple projects once. This multi-megs project is much more bearable now that you know that its dependencies will be reused by all other projects on your machine now.
+
+### D. “perfect” hoisting due to the removal of the filesystem limitations
+
+An example will be worth a thousand words: let's say you have the following dependency tree:
+
+* `top-level`
+    * `package-a`
+        * `package-c@1.0.0`
+    * `package-b`
+        * `package-c@1.0.0`
+    * `package-c@2.0.0`
+
+In this situation, even though it is found multiple times inside the dependency tree, `package-c@1.0.0` cannot be hoisted in such a way that it can be installing only once. This is because doing so would conflict with the strict requirement of `package-c@2.0.0` declared by the top-level.
+
+Since Plug'n'Play flattens the dependency tree while still preserving the links between the nodes, the paths Node will get will be the same for any `package-c@1.0.0` inside the dependency tree, causing the package to be instantiated a single time.
+
+### A. users cannot require dependencies that aren't listed in their dependencies
+
+A common problem was that it was extremely easy for a library author to start relying on a package and forget listing it inside the dependencies. Because these broken dependencies were being pulled by dev dependencies before being hoisted to the top level, they often happened to work fine in development environments and break in production.
+
+This problem isn't possible anymore with Plug'n'Play, because Yarn is aware of the whole dependency tree and can instantly decide whether a resolution request is valid or not for a given package. As an added bonus, it is able to know the exact reason why a require cannot succeed, which further improves the developer experience:
+
+```
+Error: You cannot require a package ("foobar") that is not declared in your
+dependencies (via "/Users/mael/app/some-script.js")
+```
+
+```
+Error: Package "foo@0.0.0" (via "/path/to/foo/0.0.0/node_modules/foo/index.js") is
+trying to require the package "foobar" (via "foobar") without it being listed in
+its dependencies (react, react-dom, foo)
+```
+
+### F. package instantiations now obey strict & predictable rules
+
+As mentioned in the previous point, it happens that the hoisting may not be applied fully. But the thing is, it often can be. Which makes it impossible for a project to know for sure how many times it will be instantiated by Node, and plan accordingly.
+
+The Plug'n'Play API makes it a goal to provide strong guarantees that library authors can rely on. Anything that would cause  the contract to be broken in some circumstances is unacceptable, and as a result cannot be used as guarantee.
+
+
+> **Peer dependencies: The Return**
+>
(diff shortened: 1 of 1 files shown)


orta:
I read the paper. I love the concept, moving the resolver into an exportable the `pnp.js` is a particularly smart idea.

davidnagli:
Would this have to be implemented on a package manager level?

arcanis:
@davidnagli Mostly package managers. Bundlers could however be able to consume the `.pnp.js` file to do special processing.

azz:
# ⚰️`node_modules`

This is fantastic!  

A few random thoughts:

- This will probably break a lot of tools and libraries that rely `node_modules` existing, like [`read-pkg-up`](https://www.npmjs.com/package/read-pkg-up) (7m weekly downloads). A migration strategy would need to be documented.

- How will `bin` commands be handled? Currently they're in `node_modules/.bin` which is added to the `$PATH`. Will that still exist or be moved to the cache? (I guess `PATH=$PATH:$(yarn bin)` could help in that case.) 

- Has this approach been tested with other tools that modify node resolution, like [`esm`](https://github.com/standard-things/esm)?

  ```sh
  # will this work?
  node -r esm -r .pnp.js index.js 
  ```

- > In order to solve this, Plug'n'Play details a special case when a package makes a require call to a package it doesn't own but that the top-level has listed as one of its dependencies.

  This would have to include `peerDependencies`, right?

zkochan:
For reference, I always wanted to symlink packages directly from the store in **pnpm** (related issue: https://github.com/pnpm/pnpm/issues/1001).

We also did a partial implementation (using the `--independent-leaves` flag). However, too many packages in the ecosystem currently rely on their real location, so we decided to make the change later. But it would be a lot faster than the current algos used by npm/yarn/pnpm.

Lots of packages already don't work with pnpm because of its [strict node_modules](https://medium.com/pnpm/pnpms-strictness-helps-to-avoid-silly-bugs-9a15fb306308). This is one of the reasons pnpm isn't adopted as much as Yarn or npm. I see in the RFC that it is planned to make this hooked resolution algorithm "strict" as well. I think it would be good for pnpm as we don't have the power to make the ecosystem fix itself. Even fixes that *we contribute* are sometimes not merged and published for years.

So this is a very brave design. I wonder how it will be welcomed when all the issues will arise. I wonder whether the design will be adjusted or you will insist to keep the strict resolution (this scenario would be best for pnpm)

accepted/0000-plug-an-play.md
```diff
@@ -0,0 +1,302 @@
...
+        * `package-c@1.0.0`
+    * `package-b`
+        * `package-c@1.0.0`
+    * `package-c@2.0.0`
+
+In this situation, even though it is found multiple times inside the dependency tree, `package-c@1.0.0` cannot be hoisted in such a way that it can be installing only once. This is because doing so would conflict with the strict requirement of `package-c@2.0.0` declared by the top-level.
+
+Since Plug'n'Play flattens the dependency tree while still preserving the links between the nodes, the paths Node will get will be the same for any `package-c@1.0.0` inside the dependency tree, causing the package to be instantiated a single time.
+
+### A. users cannot require dependencies that aren't listed in their dependencies
```
such: A. => E.

accepted/0000-plug-an-play.md
```diff
@@ -0,0 +1,302 @@
...
+
+The easiest solution would be for Node to implement native support for loading Plug'n'Play if detected. This is obviously not something that can be taken lightly, so this will only become viable once Plug'n'Play will have proven its value.
+
+### D. tools relying on crossing package boundaries in order to load their plugins need help
+
+Some packages try to require packages they don't directly depend on for legit reasons. Most of those are trying to do this in order to reference plugins that their users declared in their configuration. Since they don't list those plugins in their dependencies (nor should they have to), Plug'n'Play is supposed to deny them access.
+
+In order to solve this, Plug'n'Play details a special case when a package makes a require call to a package it doesn't own but that the top-level has listed as one of its dependencies. In such a case, the require call will succeed, and the path that will be returned will be the exact same one as the one that would be obtained if the top-level package was making the call.
+
+### e. edit-in-place workflows need different tools
```
gokaygurcan: e. => E.

/ping @such

mischnic:
Strange that npm has just released a similar concept (as package manager called crux): https://blog.npmjs.org/post/178027064160/next-generation-package-management

accepted/0000-plug-an-play.md
arcanis: tfw you reorder sections and forget to update the identifiers

arcanis: tfw you reorder sections and forget to update the identifiers

schmod:
My $0.02 is that any changes to replace the `require.resolve` algorithm should take place in the Node.JS core itself, and be guided by Node's community process.

An _awful lot_ of existing code and infrastructure depends on the current behavior of `require.resolve`, and will break if those semantics are changed.  The changes being discussed have the potential to fork or break the Node/NPM ecosystem, and should not be taken lightly.  

I don't want to have to worry about whether my dependencies are compatible with node_modules, pnp, or crux.

If a future version of node allows package managers to supply an alternative implementation of `require.resolve`, that's fine, *but* there should explicit guidelines about how that should work, and how developers should write package-manager-agnostic code.  This would be a semver-MAJOR change to Node.

Droogans:
How is this going to change yarn checksum behaviors, if at all? I noticed that it wasn't even mentioned.

arcanis:
@schmod Note that the goal here isn't to sidestep the Node processes. Rather, we want to show a practical implementation that we know is solid enough for a first iteration, and use it to ignite discussions.

If you want to compare it to something else, see it as what Boost sometimes is compared to the standard C++ library: it's only after `boost::thread` and `boost::filesystem` were proved to work that the WG21 felt confident enough to use them as base for the standard library, knowing that the design was of good quality not only in theory, but also in practice.

donaldpipowitch:
I'm always interested in changes like this one. 👏

I guess this would break TypeScripts `@types` resolving logic which expects these modules inside `node_modules`. What do you think about this change @DanielRosenwasser? Would you support that in TypeScript?

accepted/0000-plug-an-play.md
```diff
@@ -0,0 +1,302 @@
...
+
+### B. Installs can now be efficiently cached even on ci
+
+Now that files don't have to be copied anymore, efficient caching becomes a breeze. The actual Yarn cache can be persisted into locations shared between all CI instances, making it possible to skip installs altogether provided both the cache and the `.pnp.js` file are made available.
+
+Where some projects were spending more than two minutes running (like is the case for react-native, for example), Plug'n'Play now allows them to spend this time actually running their tests, decreasing the load on the CIs and making for a better developer experience - we all hate waiting for the tests to end.
+
+### C. users working on multiple projects across a system won't pay increasing install costs
+
+A common occurence in the Javascript world is developers working on multiple disconnected projects sharing similar dependencies (for example all projects created through `create-react-app`). Due to how package managers currently work, the files used by those projects were typically copied from the cache into multiple `node_modules`, multiplying both the total size of the installs on the disk and the time wasted running installs.
```
chengjianhua: typo: `occurence` -> `occurrence`

chengjianhua: typo: `occurence` -> `occurrence`

accepted/0000-plug-an-play.md
```diff
@@ -0,0 +1,302 @@
...
+### A. require.resolve
+
+The `require.resolve` function is problematic in that it does two things in one:
+
+* On one hand it returns an identifier that, when passed to `require`, will allow any module to load a file that would typically only be accessible from another module.
+* On the other hand it converts an identifier into a path on the filesystem.
+
+The reason this is a problem is that both of those actions don't have the same meaning and as such interfere with each other. The symlinks used for the virtual packages implementation referenced in Section 3 are a direct consequence of this: while it would be possible to implement this concept by making `require.resolve` create and return special in-memory identifiers that `require` would be able to understand, it wouldn't be possible to use those identifiers as paths (unless we were to patch the `fs` module, which is totally unacceptable).
+
+A fix would be to split require.resole in two:
```
chengjianhua: typo: `resole` -> `resolve`

boblauer:
An important note from the `crux` readme:

> You can still install things in your node_modules folder and those versions will be used in preference to the cached version. This opens a path to live-editing of dependencies (sometimes a necessary debugging technique) ...

Will this be possible with pnp as well? I poke around in node_modules very often to debug and better understand the libraries I'm using.

accepted/0000-plug-an-play.md
```diff
@@ -0,0 +1,302 @@
...
+
+* A package **MUST NOT** be able to require a package that isn't listed in its dependency detail. The dependency detail is the sum of `dependencies` and `peerDependencies` for all packages, plus `devDependencies` for the top level package if running in development mode.
+    * An exception is made if the package being required is listed in the dependency detail of the top-level. In this case, the package making the request will obtain the exact same **instance** than if the top-level package had made the require call (note the emphasis on instance rather than version).
+    * We however discourage packages from relying on this exception, since it's only been implemented to lower the adoption cost and help plugin systems. Packages should prefer using `peerDependencies` if applicable.
+* Two packages depending on the same reference of the same dependency that itself has a transitive peer dependency **MUST** get the exact same instance of this dependency, whatever their locations in the dependency tree are.
+* Two packages depending on the same reference of the same dependency that itself has a transitive peer dependency **MUST** get different instances of this dependency.
```
chengjianhua: This rule has the same condition as the previous one, but have different results. Is it a mistook?

arcanis:
> Will this be possible with pnp as well? I poke around in node_modules very often to debug and better understand the libraries I'm using.

`yarn unplug <pkg-name>` will put a copy of the specified package into `.pnp/unplugged`. You can then inspect/edit this package as you see fit, and once you're done you just have to `yarn unplug --clear-all` to go back to normal 😃

accepted/0000-plug-an-play.md
arcanis: Yes, the first one should be:

> * Two packages depending on the same reference of the same dependency that *doesn't* have any transitive peer dependencies **MUST** get the exact same instance of this dependency, whatever their locations in the dependency tree are.

In short, if two packages resolve to `lodash 1.0.0`, they are guaranteed to share the exact same 
instance (because `lodash` doesn't have any peer dependency).

If two packages depend on `react-dom 1.0.0`, they are guaranteed to each have their own instance (because `react-dom` has a peer dependency on `react`).

deepsweet:
>The current implementation overrides `Module._load`, but Node 10 recently released a new API that we plan to use to register into the resolver.

Could you please provide a link to that new API?

arcanis:
Cf [loader hooks](https://nodejs.org/api/esm.html#esm_loader_hooks). This API only applies on imports coming from ESM modules, unfortunately.

davidk01:
How does this work with modules that wrap native code? So is the idea that no more native modules will be allowed?

Aghassi:
Noticed the bit about `postinstall` going away. Agree the premise for security. I'm curious how that would impact tools like https://github.com/typicode/husky that help developer workflow, but rely on a post install task to setup. Is there a story around how these types of tools would be able to exist in this new world?

transitive-bullshit:
Really interesting proposal -- Glad to see that the `yarn` team is thinking radically towards the future!

One issue with relying on `unplugged` is that my 99% of use case for `yarn link` is to link to a version of a dependencies that I'm developing in parallel, and maintaining that symlink is important so I can edit the "real" checkout of a dependency instead of a temporary throwaway version in the `.pnp/unplugged` folder.

Aside from this issue, I'd love to hear the yarn team's thoughts on @azz's questions. I think @schmod's excellent point is answered pretty well by @arcanis's analogy to `boost`, but it would be great to understand what the Node and NPM folks think early on in this process.

@zkat I'd really love to hear your thoughts on how this proposal differs from and is analogous to some of the design decisions npm has been making with `crux`.

Thanks && I love the Node community because of awesome developments like this!

rivertam:
Also on the post-install but less agreeable.

> While native modules have their usefulness, WebAssembly is becoming a more and more serious candidate for a portable bytecode as the months pass.

This impacts a whole class of packages which genuinely rely on actually using some other language. For example, I've built a binding for a camera that interacts with linux4video2. While the camera has official bindings for both Node and C++, it doesn't have bindings for wasm. For this project, we genuinely need the low level capabilities, speed, and parallelism that C++ provides, but we can't use wasm because none of the APIs are available for wasm.

The solutions proposed in *this* proposal don't impact this in any way, as far as I can tell, and we'll be able to use pnp across all our projects. However, I'm just a little nervous reading that sentence.

saschagrunert:
Very smart and clean concept about the next generation package handling with yarn. Great! 👍

arcanis:
> The solutions proposed in this proposal don't impact this in any way, as far as I can tell, and we'll be able to use pnp across all our projects. However, I'm just a little nervous reading that sentence.

@rivertam Just to be clear: postinstall scripts will be supported, there is no plan to deprecate them in this RFC. I guess I should have been clearer 😃 

> One issue with relying on unplugged is that my 99% of use case for yarn link is to link to a version of a dependency that I'm developing locally in parallel, and maintaining that symlink is important so I can edit the "real" checkout of a dependency instead of a temporary throwaway version in the .pnp/unplugged folder.

@transitive-bullshit I have some research to do on this side - I believe `yarn link` can be made compatible with PnP without problem - we just have to do the same thing than `unplug` except that we would generate a symlink instead of a real folder. The main problem will be to reconcile both dependency tree, but I believe it can be done without too much issue 🙂 

> This will probably break a lot of tools and libraries that rely node_modules existing, like read-pkg-up (7m weekly downloads). A migration strategy would need to be documented.

@azz Yep, totally. For most packages, using `require.resolve` is enough, but some other might require a bit more tooling. From my experiments, it wasn't super common, and could be fixed without too much efforts. Since projects like Webpack and Babel happen to work just fine, I'm confident we should be mostly fine 👍 

> How will bin commands be handled? Currently they're in `node_modules/.bin` which is added to the `$PATH`. Will that still exist or be moved to the cache? (I guess `PATH=$PATH:$(yarn bin)` could help in that case.)

@azz The binaries are "hidden in the cache", but you can access them through `yarn run <bin-name>` just like before, which does the heavy lifting of figuring out which tools are provided by your dependencies.

> This would have to include `peerDependencies`, right?

@azz Not entirely sure what you mean, but this "fallback to the top-level" doesn't consider whether things are a dependency or not. If a package makes a require call to something it doesn't own (either through dependencies or peer dependencies, which are resolved at install-time).

> I might be missing something, but why have relative paths from the workspace root to the yarn cache?

@azz They were initially absolute, but I made them relative to the location of the .pnp.js file (which is at the root of the project). It could be relative to an environment variable, even if I'm not sure it's something we really want to encourage. Opinions welcome!

transitive-bullshit:
> The binaries are "hidden in the cache", but you can access them through yarn run <bin-name> just like before, which does the heavy lifting of figuring out which tools are provided by your dependencies.

I think a good sanity check for any implementation would be to make sure that `npx` still works with binaries installed locally to a package instead of fetching them remotely. imho `npx` is quickly becoming a part of the de facto standard Node.js dev's toolkit.

schmod:
> @schmod Note that the goal here isn't to sidestep the Node processes. Rather, we want to show a practical implementation that we know is solid enough for a first iteration, and use it to ignite discussions.

I appreciate the response, and hope that this proposal (and initial implementation) will be clearly marked as experimental, and that the maintainers will engage the official Node community to begin discussions about the long-term evolution of the core module loader.  

The JS community has already suffered a number of [self-inflicted wounds](https://babeljs.io/blog/2018/07/27/removing-babels-stage-presets#setting-expectations) from TC39 proposals that were widely used/implemented in the wild, before those proposals were finalized or formally adopted.  I would hope that, at a minimum, we actively encourage all developers publishing code to NPM to maintain support for `node_modules` in their applications until the community has reached a consensus about how we should move forward.

Openly experimenting with proposals like this one is a great way to collect feedback, and improve the quality of the proposal.  I only hope that we can establish some expectations and limitations upfront about how [users of] this proposal should coexist with the existing Node ecosystem.  

---

tl;dr;
* Evolution is good!  Diversity is good!  Experimentation is good!
* The community benefits from having a consistent set of expectations about how Node loads modules.  It's okay if the "official" spec sets few expectations!  
* The long-term end-goal should allow users to freely and transparently switch between npm/pnp/crux without modifying their applications.  In the interim, we absolutely must ensure that there's a seamless upgrade path for module authors and consumers.  Module authors should have guidelines about how to prepare their code for pnp/crux without breaking compatibility with the existing ecosystem.

liftM:
This is a good idea, but I think there should be a method to access the Plug 'n Play API in a language-agnostic way **without requiring a Node.JS runtime**.

At FOSSA, we do dependency analysis in a variety of environments. We need to provide results reliably across these environments, including environments where Node.JS is not available. This use case is surprisingly common for us:

1. A project-specific CI image with Node.JS does the actual `npm install` and `npm test`.
2. An organisation-wide CI image runs [`fossa-cli`](https://github.com/fossas/fossa-cli) to do dependency analysis. This image **does not** have Node.JS.

Often, we need to integrate with the organisation-wide CI image due to internal political or resourcing constraints that are out of our control. This is often the only feasible way to do integrate across thousands of projects in an organisation.

With the current `node_modules` scheme, this is easy: we emulate Node's behaviour by searching for `node_modules` folders.

The proposal for Plug 'n Play provides an executable JavaScript file as an API. This is hard for us to handle when we don't have a JavaScript interpreter available. We must either package and ship a JS interpreter with the dependency analysis tool (please don't make me write this code) _or_ convince the organisation to add Node.JS to their organisation-wide CI image (this is very hard, and the build/release/CI/devops teams at these organisations will rightly push back that this adds too much complexity and attack surface to their images).

Ideally, the public API would be language-agnostic. For us, a best case scenario is a dependency file format that is readily parse-able (e.g. `TOML`). I believe that the increase in ease-of-extensibility is worth the alleged cost in flexibility of providing a fixed file format as an API.

lhorie:
How do you deal with native ES6 module imports (i.e. `--experimental-modules`)? I was experimenting with a similar concept a while back, but couldn't find a way to hook into module resolution for ES6 imports.

elyalvarado:
Will it work with react-native libraries that include native code? In this case it isn't only executable binaries, but the actual iOS Xcode project and the Android gradle files for the project make references to the iOS and android folders in the react-native library which are located now in `node_modules`. I'm talking about non javascript files, which don't go through node resolver at all. By example, this is the actual `android/settings.graddle` file from a React Native project:

```
rootProject.name = 'TestRN'
include ':app'
include ':detox'
project(':detox').projectDir = new File(rootProject.projectDir, '../node_modules/detox/android/detox')
```

sandro-pasquali:
*rationale

arcanis:
Hey @liftM - please check the whitepaper, it includes a rational explaining why we went with a Javascript file. If you still think strongly about this after reading it, I encourage you to open an issue (on the Yarn repository, since they're not open here) and I'm open to discuss it there in more details 🙂 I'd like you to be ready to answer at least two questions if you do so:

- How significant is your use case? The flexibility you talk about comes at a price, which is the flexibility for everyone else. Is it worth it?
- Given a standardized API, what prevent you from running a script after Yarn that would consume the API and generate a file in whatever format is adaptable to your use case?

> How do you deal with native ES6 module imports (i.e. `--experimental-modules`)? 

Right now we don't, but Node 10 as [an API](https://nodejs.org/api/esm.html#esm_loader_hooks) for this, so this shouldn't be a problem. CommonJS currently is the only import mechanism that doesn't offer similar mechanisms and needs to be patched.

rightaway:
How would this work with using root-relative paths for `require` statements? I add a `node_modules` directory under my `src` directory which contains a symlink `root_dir` that points to that `src` directory, so all modules under `src` can do `require('root_dir/some_file.js')` which makes for easier require statements because they're relative to the root rather than the current file. Would this still be possible?

arcanis:
> Would this still be possible?

Yes, you'll just have to declare `root_dir` as a `link:` dependency of your top-level package.json. The following should suffice:

```
{
    "dependencies": {
        "root_dir": "link:./"
    }
}
```

jdalton:
@azz 
> Has this approach been tested with other tools that modify node resolution, like [`esm`](https://github.com/standard-things/esm)?
```shell
# will this work?
node -r esm -r .pnp.js index.js 
```

This should "just work" with `esm` _(or we, `esm` core, will make it work)_.
We support other module resolution customization/override packages like `pnp.js`.

arcanis:
@rightaway This is likely a bug - can you open an issue on the code repository and mention me there?

texastoland:
Consider the wall of identical dependencies needed to build practically anything: TypeScript and definitions, Babel and presets, ESLint and parsers/presets, Prettier and plugins, Jest and utilities, Reason (if I'm lucky) and BuckleScript testing libs, etc.

Projects like Parcel, eslint-config-shopify, and personal bootstraps these as `dependencies` so when you add `parcel-bundler` in `devdependencies` you get its transitive dependencies for free. It's effectively a hack to ship "package bundles".

Will there be any comparable facility with Plug'n'Play?

texastoland:
> _That said,_ it's not that bad, and most issues have identified solutions.

Thanks for the answer. The issue for me is duplication in `package.json` but I do understand. I'm working on a project to sync `scripts` in `package.json` with configs and presets. It could work for `dependencies` too.

AlexLandau:
When my team migrated to Yarn Workspaces, some steps in our build processes broke due to files no longer being installed at `node_modules/foo`. Some of these just became references to `../../node_modules/foo` because it is difficult in Bash to reimplement the Node.js require algorithm. These will break again if/when we move to Plug'n'Play.

To help create a more lasting solution for these cases, would it be reasonable to add a `yarn locate`/`yarn find`/`yarn resolve` command that accepts as an argument a package name and outputs the location of the package in the file system? This avoids having to install an additional command-line tool, copy-paste a function into every script, or rewrite existing scripts from Bash into Node.

azz:
@AlexLandau `node -p "require.resolve('foo')"`?

AlexLandau:
@azz I missed how the `node -r .pnp.js` would work in the proposal. That's probably an adequate solution, thanks.

texastoland:
I assume `yarn global` will keep using `node_modules`? If not can global package bins be added to `PATH`?

zindel:
Absolutely agree with @liftM's  [feedback](https://github.com/yarnpkg/rfcs/pull/101#issuecomment-421119673). I like the optimization and absence of the `node_modules`, but disagree with the `pnp.js`:

1. Non-Node based tools are totally discriminated. 
2. I don't mind the idea of "package managers dictating the resolution behaviour", but I'd rather express it using the consistent set of rules + required config and not an arbitrary piece of code to be executed at a building phase. I.e. the node resolution algorithm is [well specified](https://nodejs.org/api/modules.html#modules_all_together), while `.pnp.js` can change with the next install.

Would it be possible to split the `.pnp.js` into `.pnp.json` + `.pnp.js` which consumes it?

zindel:
@arcanis Thanks for the quick reply!

> @liftM use case was particular in that they don't want to run Node at all - something that I think isn't quite common amongst Yarn users.
To be honest that is our use case too :) We're working on the Fastpack bundler, where we try to keep the builds as fast as possible. I.e. it can take ~0.8s for the cold start and ~0.15s for the case when persistent cache is used (on the ~5Mb bundle). Spawning the node process would add up another unavoidable delay.

Anyway, I feel like I need to try integrating it and have some real numbers to compare.

shinework:
image

petetnt:
re: modules that would break under this plan. The similar [`npm` POC](https://github.com/npm/tink] includes a todo that states:

>  add fallback where "incompatible" packages get dumped into node_modules (and tagged a such in package-map.json)

Would that be an option for `yarn pnp`? Or a mechanism to mark certain dependencies incompatible?

daKmoR:
There is a somewhat similar idea of using a map to provide package names to the browser.
https://github.com/domenic/package-name-maps

As far as I can see each map stories completely different information.
So I could see the creation of such a "package-name-map" as post-install step using pnp as a source of "truth".

So as the main goal of such a package-name-maps is a flat installation for the browser, I'm wondering it should "work" combined with "yarn flat" or if it should have its own way creating the flat map (via pnp api?).

I'm just curious :)

arcanis:
> Would that be an option for `yarn pnp`? Or a mechanism to mark certain dependencies incompatible?

We already have the concept of "unplugged modules" (primarily meant for debug purposes, and triggered manually through `yarn unplug`). We'll extend this so that Yarn will automatically unplug packages with post-install scripts, as they are the most common issue people will have.

Packages with incompatibilities related to crossing package boundaries won't be subject to this automatic unplug, because it wouldn't help them: even if we were to unplug them, they would still be alone in their `node_modules`, preventing them from interacting with anything else.

And to be extremely clear: manual unplug (through `yarn unplug`) is only meant to be used for debugging purposes. I don't recommend any package from using it as a formal installation step (part of the reason why it cannot be configured through the package.json at the moment). Package authors should instead find ways to work with the package constraints they are given (which are the same one as currently, except that we now enforce them), otherwise they'll just make the developer experience of their tools worse for their users.

> So as the main goal of such a package-name-maps is a flat installation for the browser, I'm wondering it should "work" combined with "yarn flat" or if it should have its own way creating the flat map (via pnp api?).

That sounds like a prime example for a first experimentation! I've been made aware of the proposal during the PnP development, and while I still think the single-file approach is the one that'll work best for the Node ecosystem, generating package-name-maps files from the PnP API sounds super valuable. If anyone wants to give it a try, send me a mail with your results, however positive or negative!

nerumo:
how do docker image builds make use of this? the cache can't be injected during docker build time since volume mounting isn't available on docker build time (yet).

tracker1:
@zkochan IIRC symbolic links in windows requires elevated privileges.

zkochan:
@tracker1 that is why pnpm uses junctions on Windows.
If you have any questions I am ready to answer in our gitter chat.
You can find some info about how we solved issues for Window on our FAQ page: https://pnpm.js.org/docs/en/faq.html

but in a nutshell, the solution that pnpm uses for creating the non-flat node_modules works on all operating systems

geoffdavis92:
This proposal sounds great; increased performance of installation and storage is always welcome in my book.

I'm curious how this would affect the ability to inspect individual modules' contents/source in the even that I run into an issue and want to check the code:

1. Would there be additional `yarn` commands created that could "explore" the package's dependencies?
2. Could the `.pnp.js` file contain an API that could act as a sort of repl to view package contents, perhaps opening them with a arg-passed terminal text editor or with `cat` or `bat`?

millette:
FYI, a couple of deno discussions related to packaging:

* https://github.com/denoland/deno/issues/8
* https://github.com/denoland/deno/issues/288

arcanis:
> I'm curious how this would affect the ability to inspect individual modules' contents/source in the even that I run into an issue and want to check the code:

`yarn unplug` will copy a package from the cache and put it into a directory, suitable for quick exploration and patching.

> Could the `.pnp.js` file contain an API that could act as a sort of repl to view package contents, perhaps opening them with a arg-passed terminal text editor or with cat or bat?

That sounds out of scope for this proposal 🙂

geoffdavis92:
@arcanis awesome, thanks for the reply 👍

i0natan:
May I ask, probably due to my own ignorance, what's wrong with node_modules or any other way of storing all dependencies within the project?

transitive-bullshit:
@i0natan read the pdf linked in the original post 😃specifically the motivation section.

KenanSulayman:
This proposal is an applaudable effort, although I would appreciate a more generic standards process. I can already see npm Inc. trying to rebuild this feature just a lot less competent, directly competing with this and then forcing their implementation onto every single developer who didn't adopt yarn, yet.

That said, I have tried the pnp implementation provided by this PR and unfortunately have found it not to work with any of our codebases; specifically when using Node “binary”-scripts (in our case *webpack-dev-server* is used).

It seems the third-party binary being called from a script specified in package.json fails to reference dependencies of the project:

```
...
...
ERROR in multi /Users/xxx/Library/Caches/Yarn/v3/npm-webpack-dev-server-3.1.4-9a08d13c4addd1e3b6d8ace116e86715094ad5b4/node_modules/webpack-dev-server/client?http://0.0.0.0:8085 webpack/hot/dev-server ./src/index.tsx
Module not found: Error: Can't resolve 'cache-loader' in '/Users/xxx/some/directory'
 @ multi /Users/xxx/Library/Caches/Yarn/v3/npm-webpack-dev-server-3.1.4-9a08d13c4addd1e3b6d8ace116e86715094ad5b4/node_modules/webpack-dev-server/client?http://0.0.0.0:8085 webpack/hot/dev-server ./src/index.tsx
...
...
```
The `cache-loader` dependency is referenced from the webpack configuration file of the local project and loaded by the `webpack-dev-server` third-party "binary"-script, which then fails to locate the `cache-loader` dependency.

How did you solve this for the build-processes at Facebook, or if you didn't have this problem to begin with, what is the suggested approach to dealing with third-party tooling in scripts defined in package.json?

arcanis:
@KenanSulayman webpack-dev-server should work - cf the [following PR](https://github.com/yarnpkg/pnp-sample-app/pull/1/files) that adds it to the sample app. Are you sure the [Plug'n'Play loaders](https://github.com/yarnpkg/pnp-sample-app/blob/8b430a1902da5717110ac2990b55d218e0710cc3/webpack.config.js#L26-L36) are correctly configured? You also need to configure the plugin for the `resolveLoader` key, since Webpack ignores the `resolve` field when loading its own plugins.

> What is the suggested approach to dealing with third-party tooling in scripts defined in package.json?

When they use their own `require.resolve` implementation (like Webpack, which uses `enhanced-resolve`), you might need to add a plugin. For all other scripts, running them through `yarn run` and/or `yarn node` should be enough.

CrabDude:
FWIW, the `post-install` cache conflict raised in the pdf would be ameliorated by including the `NODE_MODULE_VERSION` in a cache entry's hash *if*
1. A `post-install` script ran
2. The resulting physical (filesystem) entries differ. (i.e., file content/existence)

This could be repeated for known meta-values of significance (e.g., OS version for `fsevents`). In the future, packages could even pre-declare meta-values to be considered for cache heuristics.

DanielRosenwasser:
Hey @arcanis, thanks for putting this together. I'd like to try to discuss the impact this might have on the TypeScript community. While I'll be traveling for the next few weeks, maybe we can schedule a call or something in the near future.

From my personal view, the high-level goals sound like a great idea. Apart from the use-cases given here, when it comes to powering tools like TypeScript (including editor scenarios for both TypeScript and JavaScript in tools like VS and VS Code), resolution tends to be a pretty costly process that can cause delays. Minimizing that sounds great!

However, there are a few issues I'd like to raise.

# Alternative Resolution

You may already know, but TypeScript effectively overlays a "mirrored" resolution process to find files it's interested in. Specifically:

* When resolving files without file extensions, in addition to `.js` files, TypeScript first searches for `.ts`, `.tsx`, and `.d.ts` files.
* When resolving `foo` from `node_modules`, in addition to `*/node_modules/foo`, TypeScript resolves from `*/node_modules/@types/foo`
* When resolving from `package.json`, in addition to resolving from the `main` field, TypeScript resolves from the `types` field (as well as the `typings` field)
  * In TypeScript 3.1, we will have something for version selection called `typesVersions`.

So even if TypeScript had some sort of resolution support to plug into Yarn PnP, PnP itself is oblivious to what TypeScript is actually trying to search for.

# Arbitrary Code Execution

I think @liftM covered some of this already on https://github.com/yarnpkg/rfcs/pull/101#issuecomment-421119673, but I think another broadly-applicable motivating scenario would be helpful here.

Let's say we were able to resolve this issue, and that language services that power editors were able to `require` or start up a server from the `.pnp.js` file. Ideally, this file just runs, users are happy, and they go about their lives.

Now imagine someone places a `.pnp.js` file at the root of a repository, clones it, and opens an editor there. That editor now either has the choice of executing arbitrary code or turning itself into a security notification carnival. The former is obviously not ideal, and the latter leads to a very undesirable user experience. Users either hit accept anyway, or the editor experience shuts down entirely.

# Bifurcation

Much as these ideas are great, npm's simultaneous effort on [tink](https://github.com/npm/tink) (formerly crux) means that there are potentially two approaches for us to support which isn't ideal.

# Ideas & Mitigations

I think that @sokra had some [good insights over Twitter](https://twitter.com/wSokra/status/1041972904668422144).

* A static file is definitely easier to analyze and doesn't introduce arbitrary code execution problems.
* The complete mapping of files in packages makes it fully possible for any tool to overlay its own resolution process which is crucial to TypeScript.
* Having a tool that can actually be imported or spawned as a server/daemon to understand this new feature is absolutely a great idea to ensure tools like Flow can still work without a full reimplementation.

arcanis:
> So even if TypeScript had some sort of resolution support to plug into Yarn PnP, PnP itself is oblivious to what TypeScript is actually trying to search for.

The Plug'n'Play API is split in two parts: `resolveToUnqualified`, and `resolveUnqualified`. The first one is the static resolution that converts `lodash/foo` into `/path/to/cache/lodash-1.2.3/foo`. The second one is the one that converts `/path/to/cache/lodash-1.2.3/foo` into `/path/to/cache/lodash-1.2.3/foo/index.js`.

So in your case, your resolver would just have to use `resolveToUnqualified` in order to get the basic path, which you would then continue resolving as you see fit. You can see an example on the [webpack resolver](https://github.com/yarnpkg/pnp-sample-app/blob/master/scripts/webpack-resolver.js#L31), which uses `resolveToUnqualified` then defers to the rest of the `enhanced-resolve` plugins to compute the rest of the resolution.

> Now imagine someone places a `.pnp.js` file at the root of a repository, clones it, and opens an editor there.

It's a great point, I didn't consider it before. It's worth noting that this problem already occurs now, though: while not native to the editor, various extensions already transparently execute Javascript files obtained from freshly cloned projects - a classic example being the eslint plugin for vscode.

accepted/0000-plug-an-play.md
```diff
@@ -0,0 +1,302 @@
...
+
+## 2. Motivation
+
+When the first Javascript package managers appeared, they had to use the tools available to them. As a consequence, they were implemented on top of the Node resolution algorithm, copying the packages in such a layout that Node would be able to find them without external tool. This allowed the ecosystem to thrive, which interestingly revealed some of the scalability flaws in this approach:
+
+* The Node resolution algorithm isn't aware of what packages are, and as a result doesn't know anything about dependencies either. When a module makes a require call, Node simply traverses the filesystem until it finds something that matches the request, and uses it.
+* Installations take time, and the time required to copy files from the package managers' caches to the `node_modules` folders is one of the main bottleneck as it is heavily I/O bound, which cannot be easily optimized. Copy-on-Write filesystems can alleviate this issue to an extent but come with their own set of drawbacks, such as a lack of global support.
```
oguzzkilic: typo: managers' => managers

AlexLandau: This is not a typo. This is a use of the apostrophe to indicate a plural possessive, i.e. "the caches of the package managers".

(While I'm talking about this anyway: "without external tool" in line 23 should be "without external tools", or perhaps "without an external tool".)

catamphetamine:
The serif font on the paper makes it unreadable

kitze:
Someone gets rid of node_modules, the largest folder in your project, your computer, your neighborhood, and the universe, and YET, someone complains about A FUCKING FONT. That's our community ladies and gentleman.

arcanis:
I appreciate the support, but please let's keep it civil. Everyone is entitled to their opinions, especially on a process that specifically asks for feedbacks - 'voting' with your reactions is the best way to signal whether you agree with them or not 🙂

ChuksFestus:
ok i don't know what i did wrong but i still get node_modules :( 
i ran `npx create-react-app myapp --use-pnp` still got node_modules

ChuksFestus:
@levrik you can't really do `yarn create-react-app` but anyways create-react-app uses yarn by default if its installed so when i ran `npx create-react-app` it ran yarn

arcanis:
Hey @chulksfestus, can you open an issue on the [Yarn repo](https://github.com/yarnpkg/yarn) if you still have an issue? 🙂

It's likely caused by a wrong version of Yarn being used, and I prefer to keep this thread focused on the design of the feature, keeping the implementation out of the RFC repository. Thanks!

arcanis:
And actually, since the feature has already landed in Yarn 1.12 (currently in RC), I'm going to go ahead and merge this PR 🎊

Thanks everyone for the feedback you provided, this is but the beginning and we'll keep working with you to improve Plug'n'Play and make it a rock-solid foundation for your projects 👊

In case you have any question, feel free to open an issue on the Yarn repo or to ping me on Twitter and I'll make sure to come back to you. Cheers!

afzalah:
@arcanis why yarn doesn't simply embed/use [@verdaccio](https://github.com/verdaccio/verdaccio) and support in their development instead.

jsg2021:
@afzalah This isn't a private registry. This simply changes how modules are locally resolved & fetched.

atishaybaid:
@arcanis  To experiment with pnp,i upgraded my yarn version  to v1.12.0,and added the config in package.json
`"installConfig": {
    "pnp": true
  }`,
and `pnp.js` file got generated when running yarn.
however my build got failed, as it was not able to resolve several packages,like react,lodash

is there anything else i am missing in the configuration.?


screen shot 2018-10-09 at 7 57 43 pm

_

orta:
Please don't use this thread for support requests, use the yarn repo or stack overflow. Notifications go out to 45+ people who were giving feedback on the RFC and 60+ folks interested in RFCs for yarn.

PR #103, opened 2018-09-28
Add a proposal to deal with peer dependency noise.

alloy:
⚠️ ⚔️

### DIFF
--- a/accepted/0000-deal-with-peer-dependencies.md
+++ b/accepted/0000-deal-with-peer-dependencies.md
@@ -1,0 +1,154 @@
+- Start Date: 2018-09-28
+- RFC PR: (leave this empty)
+- Yarn Issue: (leave this empty)
+
+# Summary
+
+Allow a developer to deal with peer dependency warnings, without the need to
+install all dependencies, thus reducing warning fatigue and drawing attention to
+new warnings.
+
+# Motivation
+
+[Peer dependencies][peer-deps] are a useful way of informing the developer that
+a dependency relies on the host project to provide a transitive dependency
+required by the dependency, without imposing that transitive dependency onto the
+project’s dependency graph on its own.
+
+However, there are legitimate cases when the developer may want to ignore a peer
+dependency warning. Such as the host not being compatible with the dependency,
+the version constraint being too restrictive, or otherwise not being interested
+in the dependency.
+
+While treating warnings as errors is often useful, warnings are specifically not
+errors, because they could possibly be ignored after judgment by the developer.
+
+Currently, the inability to deal with peer dependency warnings, other than
+installing the required package/version, leads to warning fatigue, will be
+ignored, and will introduce bugs.
+
+_Regarding a version constraint being too restrictive, this usually comes up
+when the transitive dependency’s version is lower than v1.0.0 and the dependency
+has a version requirement on it using a semantic-version operator such as `^`.
+E.g. `graphql@0.13.0` will not satisfy a requirement like `graphql@^0.12.0`._
+
+# Detailed design
+
+Consider the following peer dependency warnings shown on `yarn install`:
+
+```
+warning " > @artsy/palette@2.18.0" has unmet peer dependency "react-native@^0.55.4".
+warning "@artsy/palette > styled-reset@1.4.0" has incorrect peer dependency "styled-components@>=4.0.0".
+```
+
+When the developer has determined that they do not want or cannot install these
+right now, they add these to the `ignoredPeerDependencies` list of their
+`package.json` manifest.
+
+For instance, as the project the developer is working on is a web project, they
+want to always ignore a warning about the `react-native` peer dependency of
+`@artsy/palette`:
+
+```json
+{
+    "ignoredPeerDependencies": {
+        "@artsy/palette": ["react-native@*"]
+    }
+}
+```
+
+The `styled-components` dependency of the transitive `styled-reset` dependency,
+however, is one that the developer cannot update to right now and so, _after_
+having verified that the version that they can use works for their use-case, the
+developer adds this warning to the ignore list:
+
+```json
+{
+    "ignoredPeerDependencies": {
+        "styled-reset": ["styled-components@>=4.0.0"]
+    }
+}
+```
+
+Afterwards, running `yarn install` shows no more warnings, until either:
+
+* any dependency introduces a _new_ peer dependency
+
+  ```
+  warning " > react-tracking@5.3.0" has unmet peer dependency "babel-runtime@^6.20.0".
+  ```
+
+* _another_ dependency relies on `react-native`
+
+  ```
+  warning " > @artsy/emission@3.42.0" has unmet peer dependency "react-native@^0.55.4".
+  ```
+
+* the `styled-reset` dependency changes its requirement on `styled-components`
+
+  ```
+  warning "@artsy/palette > styled-reset@1.5.0" has an updated ignored peer dependency "styled-components@>=3.9.0".
+  ```
+
+In each of these cases, the developer’s attention should be drawn to the warning
+so they can deal with it at the time they have the right context.
+
+An additional warning could be shown when a listed (transitive) dependency no
+longer exists in the dependency graph.
+
+```
+warning "styled-reset" with ignored peer dependencies is no longer depended on.
+```
+
+Finally, at any given time all warnings should be visible by passing the
+`--verbose` flag to `yarn install`.
+
+# How We Teach This
+
+This change is only additive and does not change any existing behavior of Yarn
+or how developers are supposed to interact with it. The new terminology clearly
+communicates the intent and only introduces a negated version of existing
+terminology.
+
+Its communication should follow that of Yarn’s
+[`resolutions` feature][yarn-resolutions], in that it is an extra way that a
+developer can gain control over their dependency graph and ensure its stability.
+
+Links to its documentation could be added to the existing documentation on [peer
+dependencies][yarn-peer-deps], but otherwise does not need to be altered.
+
+# Drawbacks
+
+It’s one more configuration that needs to be documented and understood.
+
+An alternative that does not introduce a new `package.json` key could be
+preferable, but that does not appear to be feasible while maintaining
+compatibility with npm.
+
+# Alternatives
+
+Introducing a new version requirement negation operator, for example:
+
+```json
+{
+    "peerDependencies": {
+        "styled-reset": "styled-components@!4.0.0"
+    }
+}
+```
+
+However, this breaks compatibility with npm and, more importantly, doesn’t allow
+for multiple peer dependencies of the (transitive) dependency to be ignored.
+
+# Unresolved questions
+
+I don’t feel strongly about the configuration, except that I dislike more
+configuration. Any suggestions that could lead to the same outcome with less
+new configuration would be greatly preferred.
+
+Perhaps the alternative design listed above could piggy-back on the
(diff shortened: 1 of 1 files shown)


alloy:
@bestander Could you tell me what/if I should be doing next to receive feedback on this?

arcanis:
Sorry, I missed your RFC - we've been discussing a similar feature (optional peer dependencies) here: https://github.com/yarnpkg/yarn/issues/6487

alloy:
While I do think having an explicit way to mark dependencies as optional is a good thing to have, that issue seems largely orthogonal to mine.

My proposal is not about labelling peer dependencies of a package as optional or dealing with side-effects of implementation details (of PnP), but rather a way for a consuming project to make decisions based on warnings and deal with them. This is something that, from my perspective, is needed regardless of the other issue being implemented or not.

Having said that, my proposal would probably allow for dealing with the PnP issue in a more timely manner, because even after adding a way for packages to list optional peer dependencies it’s going to take a while for all packages to use them. (Some packages may never use the feature while actually having optional peer dependencies.)

Off-topic: I’m slightly surprised about that other issue not being discussed as an RFC. Did I misinterpret what types of things should be discussed as RFCs?

Gudahtt:
If we decide to add an `optionalPeerDependencies` field or support an custom version range syntax for optional `peerDependencies` ([as discussed here](https://github.com/yarnpkg/yarn/issues/6487#issuecomment-428622540)), would we still want this new field as well?

I'm leaning toward yes, because library authors won't all make use of this hypothetical new Yarn-specific feature. Ignoring peer dependency warnings would still be useful for application authors that use libraries that are "misbehaving" either by declaring an overly-restrictive version on a `peerDependency` or by including a `peerDependency` that isn't required at all.

alloy:
> I'm not convinced by the example given of the version constraint being too restrictive. The library author has the ability to make whatever restriction they want on the version of a peer dependency. In that situation, typically I would think that the library should be updated or forked, rather than this warning be ignored.

In theory I absolutely agree with you. In reality, however, I already maintain so many forks and try to get those in upstream, that sometimes you need to pick your battles. Even if I wanted to, in more than one case in the past have I had trouble getting a library maintainer to release a new patch release of an older major/minor line (there may be a good reason for me to run an older version).

Ultimately, what it boils down to for me is that while I absolutely would like all version requirements to be perfect even going forward, the reality is that there will always be reasons for me to make a different decision for my project. So when I have verified that a certain version works absolutely fine with a different transitive dependency then imo I should ultimately be in control over my workspace and be able to call it a day.

> "peer dependencies" are intended for mandatory dependencies. Usually when I see a peer dependency warning that is safe to ignore, it's because the peer dependency is optional.

I didn’t even know they were strictly mandatory. I have definitely used peer dependencies as optionals, in rare cases. TIL, thanks.

Nonetheless, yeah they absolutely are being used so by more people than me right now and I think there should be a way to express that separately, so I’m :+1: on https://github.com/yarnpkg/yarn/issues/6487.

(It’s really unfortunate that `optionalDependencies` is being used for those whose installation is precarious.)

alloy:
> If we decide to add an `optionalPeerDependencies` field or support an custom version range syntax for optional `peerDependencies`, would we still want this new `ignoredPeerDependencies` field as well?
>
> I'm leaning toward yes, because library authors won't all make use of this hypothetical new Yarn-specific feature. Ignoring peer dependency warnings would still be useful for application authors that use libraries that are "misbehaving" either by declaring an overly-restrictive version on a `peerDependency` or by including a `peerDependency` that isn't required at all.

Agreed. This goes back to “[I should] ultimately be in control over my workspace and be able to call it a day”.

Put otherwise, this field is for app/project authors to override based on their informed needs, not for library authors.

Gudahtt:
>I didn’t even know they were strictly mandatory. 

Well.... they're intended to support plugins, which is a situation in which the peer is required. *A lot* of people use them for optional packages as well, so I probably worded that too strongly. I don't know what the authors of the original implementation thought about optional peer dependencies, just that it doesn't deal with them very well right now.

> Put otherwise, this field is for app/project authors to override based on their informed needs, not for library authors.

That's a great way of explaining it, thanks.

alloy:
(Btw, I now realise that I never stated that I’d love to work on this, but didn’t want to start before this was agreed upon and how it would work. Updated the OP.)

arcanis:
> I didn’t even know they were strictly mandatory. I have definitely used peer dependencies as optionals, in rare cases. TIL, thanks.

Peer dependencies are not strictly mandatory - it's just that, if missing, we'll print a warning. We won't crash the install or anything - instead it will throw at runtime if accessing the peer dependency without guard. If you have a guard, then it's perfectly fine.

The warning unfortunately leads to suboptimal UX for packages that have optional integrations with others (), so we're looking into ways to allow packages to signal Yarn that the warning is unnecessary.

>  Ignoring peer dependency warnings would still be useful for application authors that use libraries that are "misbehaving" either by declaring an overly-restrictive version on a peerDependency or by including a peerDependency that isn't required at all.

I think the best feature would be to rework the logging to assign codes + data to the various warnings, and allow our users to select which ones then want to disable from the yarnrc. I'm not fond of anything that would only work for the peer dependencies case.

arcanis:
I disagree - the behavior when a peer dependency is missing is clearly defined, so "you can't rely on anything working" isn't quite true. You can rely on **not** being able to access a missing peer dependency it (ideally for regular installs, enforced with PnP), and you can rely on being able to access any of your other dependencies. It's not an UB that could potentially mess your whole tree.

Whether `npm ls` or `yarn check` exit nonzero doesn't really matter - both of those commands have a different semantic than an install, and can afford to be stricter and exit with non-zero codes when a regular install wouldn't (whether they *should* is a different story).

Gudahtt:
> I think the best feature would be to rework the logging to assign codes + data to the various warnings, and allow our users to select which ones then want to disable from the yarnrc. I'm not fond of anything that would only work for the peer dependencies case.

I don't think storing this configuration in `yarnrc` would be a great idea. Ignoring peer dependency warnings should be a project-specific configuration, whereas `yarnrc` is meant for user-specific configuration.

As for how to ignore the warnings, the information we'd want from the user to decide which peer dependencies to ignore is exactly as described by this RFC. We'd want to know which dependency had the peer dependency, and what that peer dependency was exactly (including the version range). Anything less than that, and we would end up silencing warnings we shouldn't.

That is, the configuration would be specific to peer dependencies, and can't be generalized to fit other situations in which you might want to ignore warnings.

Though, if you do think that ignoring warnings on a per-project basis might be desired in other situations and you want to group the configuration together, we could consider something like this instead:

```
{
    "ignoredWarnings": {
        "peerDependencies": {
            "styled-reset": ["styled-components@>=4.0.0"]
        }
    }
}
```

I don't know of any other examples of warnings that should be ignored for a project though. Usually warnings indicate a problem that can be fixed.

arcanis:
> Ignoring peer dependency warnings should be a project-specific configuration, whereas `yarnrc` is meant for user-specific configuration.

The yarnrc is meant for both. If you have multiple yarnrc in your filesystem hierarchy, they'll all be loaded, from the root one to the closest one. It's a technic that's really handy precisely to define project-level configuration settings, such as the custom registry or a custom Yarn binary (we actually recommend using it to enforce the Yarn version a given team will use, regardless of their global one).

> As for how to ignore the warnings, the information we'd want from the user to decide which peer dependencies to ignore is exactly as described by this RFC. We'd want to know which dependency had the peer dependency, and what that peer dependency was exactly (including the version range). Anything less than that, and we would end up silencing warnings we shouldn't.

Yep, that's why I mentioned adding metadata to the warnings. So for example, silencing the peer dependency warning could look like this:

```
ignore-warnings:
  missing-peer-dependency:
    - name: typescript
      expected-by: ts-node
```

Gudahtt:
That all seems pretty reasonable!

Going forward, I suppose `yarnrc` should be preferred for new project configuration over `package.json`? It seems safer to add things to `yarnrc` - less likely to break existing packages that might rely upon `package.json` configuration.

alloy:
I like that, moving this to `yarnrc`, it makes the point of this being for project authors, as opposed to library authors, much clearer.

arcanis:
> Going forward, I suppose `yarnrc` should be preferred for new project configuration over `package.json`? It seems safer to add things to `yarnrc` - less likely to break existing packages that might rely upon `package.json` configuration.

Yup - the main recent exception was the recent `installConfig.pnp` flag, which I put in the `package.json` to encourage interoperability with other package managers, but overall the yarnrc is a good place to keep all project-level settings 🙂

Gudahtt:
I have a couple more questions.

```
ignore-warnings:
  missing-peer-dependency:
    - name: typescript
      expected-by: ts-node
```

That format looks pretty good, but it is missing the version range. Including the range would be useful in the case where the warning is appearing due to a version mismatch. Then Yarn could show the error again if the expected version changed, which might be helpful.

The version range would be pretty useless for the case where the peer dependency is totally optional though. If you never want to use a peer dependency, you might not care what version range is being asked for.

So perhaps the range should be optional.

Second question: should this configuration be exposed via the CLI? The [yarn config](https://yarnpkg.com/en/docs/cli/config) command is key-value oriented, which doesn't work well here. 

Maybe we could add a new subcommand for ignoring warnings (e.g. `yarn config ignore-warning <type> [--name <name> --expected-by <expected-by>]`. Or we could add an interactive prompt. That does seem a bit complex though, and not strictly necessary. I did want to bring it up because this might be the first `yarnrc` config that can't easily be specified via `yarn config`.

alloy:
> That format looks pretty good, but it is missing the version range. Including the range would be useful in the case where the warning is appearing due to a version mismatch. Then Yarn could show the error again if the expected version changed, which might be helpful.
>
> The version range would be pretty useless for the case where the peer dependency is totally optional though. If you never want to use a peer dependency, you might not care what version range is being asked for.
>
> So perhaps the range should be optional.

This is what I actually presumed was meant, but it’s indeed good to have it explicit. I do think we should be able to specify the version range.

alloy:
Sure thing 👍 Will do this next week when I’m back home.

epmatsw:
If we've got configuration setting to ignore peerDependency warnings, would it make sense to add a similar option to make peerDependency warnings present as errors instead? If we have this outlet to ignore them, making unexpected/un-ignored incompatibilities fail loudly until ignored seems like something that would be nice to have.

arcanis:
Similarly, I think I'd be more in favor of a `-Werror`-like option where warnings would be treated as errors. Would be more generic than `--enforce-peer-dependencies` or similar 😃

alloy:
Alright, updated as per the discussion and added a few more unresolved questions.

alloy:
Hey all, no rush on my side –I have plenty to do– but just wondering what the next step is so that I’m not accidentally waiting on nothing?

accepted/0000-deal-with-peer-dependencies.md
```diff
@@ -0,0 +1,191 @@
...
+
+It’s one more configuration that needs to be documented and understood. On the
+other hand, it is likely that the `ignore-warnings` project configuration will
+be expanded upon in the future, such as ignoring a missing license key in a
+dependency’s `package.json` file, and as such will be less of a niche
+configuration.
+
+# Alternatives
+
+A previous itertation of this RFC suggested the addition of a key to the
```
dcleao:
```suggestion
A previous iteration of this RFC suggested the addition of a key to the
```

alloy: Thanks!

PR #105, opened 2018-10-30
Optional Peer Dependencies

arcanis:
Related discussion: https://github.com/yarnpkg/yarn/issues/6487

cc @edmorley @jscheid @Gudahtt who were part of the initial discussion

Also cc @zkochan (pnpm), @iarna (npm), since you'll likely be affected by this change

### DIFF
--- a/accepted/0000-optional-peer-dependencies.md
+++ b/accepted/0000-optional-peer-dependencies.md
@@ -1,0 +1,52 @@
+- Start Date: 30 Oct 18
+- RFC PR: https://github.com/yarnpkg/rfcs/pull/105
+- Yarn Issue: n/a
+- Champion: Maël Nison (@arcanis - [Twitter](https://twitter.com/arcanis))
+
+# Optional Peer Dependencies
+
+## 1. Which Problem Does It Solve
+
+> I'm a library author. I want my package to provide an optional integration with some other library, but I don't want to have to ship it.
+
+The current solution to the problem described above is peer dependencies. A package will list the optional dependency as peer dependency, meaning that it's to the package consumer to also provide the library that should be used.
+
+While it currently already works, it causes semantical issues: are peer dependencies meant to be optional or not? Should a missing peer dependency cause a warning? An error? Nothing at all, and expect the runtime to check that they are available?
+
+## 2. Detailed Design
+
+Yarn will introduce a new protocol, `optional:`. If a peer dependency cannot be found but the requested range begins with `optional:` then Yarn will silence any warning and assume that the package will check at runtime whether it should use it or not.
+
+**Important:** The `optional:` protocol will not be valid in anything else than peer dependencies. We reserve the right later to potentially implement it for other fields, but for the time being we're only interested in the `peerDependencies` use case.
+
+## 3. How We Teach This
+
+- We can just call the optional protocol.
+
+- Being stricly backward compatible, it doesn't require us to push changes onto our users, so the teaching should be simple and spread on the long time.
+
+## 4. Drawbacks
+
+- It requires adoption from other package managers, otherwise there's a risk to fracture the ecosystem if they decide to go with (for example) naming the field `opt:`.
+
+## 5. Alternatives
+
+- We could introduce an `optionalPeerDependencies` key.
+
+  - It would be one more key to teach users.
+
+  - This wouldn't be backward compatible - most package managers would have no idea what to do with such a key and would ignore it entirely, breaking the tree because of the hoisting.
+
+  - Semantic isn't clear if a dependency is both in `peerDependencies` and `optionalPeerDependencies`.
+
+  - It would have a completely different meaning from the already existing `optionalDependencies` field.
+
+- We could overhaul how the dependencies are defined, and fix them once and for all.
+
+  - This sounds a huge undertaking for a problem relatively minor at the moment.
+
+  - It seems unlikely we can reach a consensus in a reasonable timeframe.
+
+## 6. Unresolved Questions
+
+Will all the package managers work properly when seeing a `peerDependencies` range prefixed by `optional:`? Logic would say that yes, since peer dependency ranges are typically an hint rather than a strict enforcement (except maybe for npm 1 & 2), but it might trigger a bug in some versions of some package managers.


zkochan:
It seems to me that the current `peerDependencies` field behaves in an "optional" way already. A warning is printed but installation does not fail.

(pnpm currently has a config called [strict-peer-dependencies](https://pnpm.js.org/docs/en/pnpm-install.html#strict-peer-dependencies) that makes installation fail if a peer dependency is not from the wanted range or is missing.)

So I think it may be solved the other way around. We think about the current "peerDependencies" as "optional" (and change the warnings to info messages) + we introduce a `strict:` protocol or a "strictPeerDependencies" field

arcanis:
> It seems to me that the current peerDependencies field behaves in an "optional" way already. A warning is printed but installation does not fail.

Yep totally - this is more from a UX perspective. We currently print quite a lot of warning, which:

1. Makes it impractical for package authors to add integrations with other tools (because it will print non-actionable warnings in their users interfaces)
2. Causes the users to disregard all warnings because there's so many of them, most of them are unactionable, and they have no easy way to figure which ones are

> we introduce a `strict:` protocol or a "strictPeerDependencies" field

That's an interesting idea - maybe it would be a good thing, yeah. I'm still not fond of the `strictPeerDependencies` field, though, since older package managers (even Yarn itself) wouldn't know how to interpret them (they would simply ignore them). That seems a bit risky, don't you think?

zkochan:
It may be done in a backward compatible way. For instance, via a `strictPeers` field that will be just an array of the strict peer dependencies.

```json
{
  "peerDependencies": {
    "foo": "1.x.x",
    "bar": "1.x.x"
  },
  "strictPeers": ["foo"]
}
```

arcanis:
I guess .. it seems a bit too convoluted imo (especially since it doesn't match the other keys names - for example `bundleDependencies` has a similar design but a totally different name).

Plus, what if we need other such "flags" later on? I'd like to avoid `devPeers` (just an example on the top of my head) or similar?

zkochan:
probably a stupid idea... but this would also work, checking if the version spec starts with `||`. and it would be backward compatible

```
{
  "peerDependencies": {
    "foo": "|| 1.x.x",
    "bar": "1.x.x"
  }
}
```

arcanis:
Hm could work - we should check that it doesn't make current package managers crash with "invalid semver pattern" or something like that 🤔

zkochan:
It should work because semver handles it fine: https://runkit.com/embed/qjynj6em684l

ljharb:
They're not optional because `npm ls` exits nonzero; the only reason they don't fail installation is that they're not installed by default. In npm < 3, they *would* install by default, and an installation failure would fail installation.

ljharb:
In other words, "required" doesn't mean "ensures the installation succeeds", it means "produces a valid dependency graph". Missing peer deps is an invalid graph.

arcanis:
@ljharb I think you also made this point [here](https://github.com/yarnpkg/rfcs/pull/103#issuecomment-429668277), right (to provide more context and discussion on the subject)?

As before, I don't really see `npm ls` as a standard as much as any other command from any other package manager, so I would tend to rather look at the user expectations rather than that. Ideally we would look at the spec, but there's none, which is in part why we're here: to clearly spec the expectations of a peer dependency being missing.

From what I saw on various threads, most users don't expect missing peer dependencies to cause their installs to fail, so I would be quite strongly opposed to making it an error. Keeping it a warning if missing seems to me the best solution, especially since it's trivial to any of us to add a flag that makes warnings into errors (think `-Werror`).

ljharb:
Most users also wouldn't expect a package's "engines" declaration having a nonmatching version of node to make their installs fail, but yarn does that.

arcanis:
> I would argue that npm install not running npm ls afterwards is an omission that masks many real problems in developer's graphs; I've caught tons of bugs in packages I maintain by running it in CI after an install.

Coincidentally I've opened a separate RFC this morning to get rid of `yarn check` in Yarn 2 (the main rational behind it being that if our users have to rely on `yarn check`, it's that we've done something wrong in `yarn install`) 😄 

> because they'd end up with valid dependency graphs that meet the requirements package authors declare for their packages.

The problem I have with this is that you're making assumptions regarding the intents package authors had when writing their packages (those that have already been published). It seems dangerous to me for a very little gain. Think of the worst case: a peer dependency *from a transitive dependency* is missing and causes the whole install to fail in a very non-actionable way.

I prefer to be a bit more conservative on this topic (even though you should know by this point that I *love* enforcing strictness, after my work on PnP) and have a forgiving behavior. After all it's not entirely the users fault if they weren't properly taught what behavior was to be expected.

ljharb:
That's intentional. An unmet peer dep in a transitive dependency *should* fail the entire install; it's an error in whatever intermediate dep included the one with the peer, and failed to either redeclare the peer, or declare a dep. It's quite actionable for the top-level user tho - peer dep requirements hoist, so all they have to do is install the missing peer dep at the top level.

arcanis:
> An unmet peer dep in a transitive dependency should fail the entire install

You can say it "should" but I'd prefer practical arguments. Whether you (or I) think it should or not isn't the point here; instead we need to find a way to work with the ecosystem in a way that:

1. can be teached
2. doesn't require a lot of changes

The fact is that some packages are already using peer dependencies to mark optional dependencies *because they don't have the choice*. Meaning that if you go your way, you would for sure break packages. In contrast, package managers don't currently fail the install if a peer dependency is missing, meaning that it's 100% sure we wouldn't break anything going this way.

I wouldn't necessarily disagree with you if we were to create a new ecosystem. In this situation, though, we only have so many jokers and I don't think it's wise to use one of them on this when we have clear alternatives.

> It's quite actionable for the top-level user tho - peer dep requirements hoist, so all they have to do is install the missing peer dep at the top level.

Eh, this one is fun and would worth a separate discussion. Long story short, this is only by accident. If a package A depends on a package B that has a peer dependency C, then A **must** provide C in some capacity (it can be through a peer dependency itself - the point is: it has to be provided).

This semantic is a bit borderline (again, it never got taught one way or the other), but I'm not aware of any popular package using such patterns so I'm not too concerned about enforcing it (note that it's off-topic for this discussion; please open a new thread if you want to discuss it). For those that would do, they would actually be helped a lot if they had access to optional peer dependencies.

arcanis:
> It should work because semver handles it fine: https://runkit.com/embed/qjynj6em684l

Wow .. I guess it should, yep. Apparently, empty string is a valid semver range 😅 Let's see what it gives with the various syntaxes proposed:

```json
{
    "peerDependencies": {
        "typescript": "optional:^3.0.0",
        "webpack": "^4.0.0"
    }
}
```

```json
{
    "peerDependencies": {
        "typescript": "^3.0.0",
        "webpack": "strict:^4.0.0"
    }
}
```

```json
{
    "peerDependencies": {
        "typescript": "| ^3.0.0"
        "webpack": "^4.0.0"
    }
}
```

ljharb:
Most every eslint plugin/config and react ecosystem package uses peer deps in this way - to explicitly declare compatibility with the “main” package (eslint, react). Webpack as well, i believe.

What I’ve never seen is people using peer deps to mean “optional” - in that case (like enzyme 2 did) they do a try/catch require and document the package in their readme. Do you have specific examples of packages that intend this semantic?

arcanis:
I can definitely look for a more comprehensive list, but one of them that triggered this issue is https://github.com/waysact/webpack-subresource-integrity/issues/90. Another one is https://github.com/TypeStrong/ts-node/issues/697.

> in that case (like enzyme 2 did) they do a try/catch require and document the package in their readme.

It's unfortunately unsafe, they shouldn't do this (hence the issues I've opened). Even disregarding PnP, peer dependencies have specific mechanisms to prevent hoisting from being made in a way that would break the require contract. By not listing the dependency, they run the risk of being hoisted in such a way that they won't be able to access the instance from their parent anymore. For example:

```
- A
    - B (optional peer dependency on C, omitted)
    - C@1
- C@2
```

In this context, the package manager will quite likely decide to hoist B to the top-level (which won't be possible for C, since the versions would clash), meaning that the instance of C it will get won't be the same one as the one used by A.

ljharb:
I agree that optional peer deps are a useful feature; what I’m arguing is that in practice they are not optional, and are not used that way.

arcanis:
So - what syntax do we go with? I'd tend to go with either `optional:` or `strict:` since they appear more readable and less confusing to me, but I'm not strongly opposed to `|` either.

Regarding optional vs strict, I think using optional would be better. I agree with @ljharb that *most* packages expect peer dependencies to be required (but disagree that *all* of them do so), and it seems better to me to warn too much than not enough.

---

Last call for the npm folks (@iarna), if you wish to contribute. It's a simple but important change and it would be better if all package managers were aligned in this. Otherwise the sheer demand will prompt us to implement it nonetheless, but it might make your devx slightly less good 🙂

rwjblue:
> Regarding optional vs strict, I think using optional would be better. I agree with @ljharb that most packages expect peer dependencies to be required (but disagree that all of them do so), and it seems better to me to warn too much than not enough.

I strongly agree here.

> Last call for the npm folks (@iarna), if you wish to contribute. It's a simple but important change and it would be better if all package managers were aligned in this. Otherwise the sheer demand will prompt us to implement it nonetheless, but it might make your devx slightly less good 🙂

I'm pretty concerned with moving forward without _some_ feedback from the NPM side. Without support from NPM package authors are really put in a bad spot where they must choose to _only_ support Yarn consumers. 

> So - what syntax do we go with? I'd tend to go with either optional: or strict: since they appear more readable and less confusing to me, but I'm not strongly opposed to | either.

If we are unable to agree on the semantics amongst package managers, IMHO we should strongly prefer syntax that at the least doesn't **break** one of them (@zkochan's suggestion of `||`)...

arcanis:
> Without support from NPM package authors are really put in a bad spot where they must choose to only support Yarn consumers.

Note that changes we make here will work on all package managers, even those that don't implement `optional:`. This is because peer dependency ranges are merely an hint, and even if they don't match the install shouldn't fail (it should warn, though). So even in the worst case scenario, the npm users will only get a warning telling them that `optional:^1.0.0` isn't satisfied even when provided `1.0.0`. Annoying but not ecosystem-breaking.

I'd still prefer for them to manifest but I can't force them to unless they want to, and given that some of them blocked me on Twitter I'm not overly confident they're in a collaborating mood. Hopefully I'm wrong there 🙂 

> If we are unable to agree on the semantics amongst package managers, IMHO we should strongly prefer syntax that at the least doesn't break one of them

To be clear about the effect (I made two packages to test: `optpeer1` and `optpeer2`):

package manager | dependency | not provided | provided
--- | --- | --- | ---
yarn | `optional:^1.0.0` | warning: missing | warning: incorrect
pnpm | `optional:^1.0.0` | warning: incorrect | warning: incorrect
npm | `optional:^1.0.0` | warning: missing | warning: missing
yarn | `l ^1.0.0` | warning: missing | ok
pnpm | `l ^1.0.0` | warning: incorrect | warning: incorrect
npm | `l ^1.0.0` | warning: missing | ok

(replace the `l` with a pipe; it broke the formatting otherwise)

I guess that `| something` might have the best forward compatibility, but I find it confusing nonetheless.

rwjblue:
Thank you very much @arcanis your breakdown with actual tests was very helpful, sorry for jumping to conclusions RE: using `optional:` would break in NPM (I'm likely still smarting RE: `link:` / `file:` differences between the two 😝).

ljharb:
Instead of syntax, why not a new top level optionalPeerDependencies key? That would both not interfere with npm, and not cause new warnings to show up in npm ls - which many packages run in CI, which means the optional syntax would indeed break CI for non-yarn users.

ljharb:
I guess I'm confused. If you decide to treat `peerDependencies` as "required" - all older yarns, and all npms, would continue to treat those as such (with the current state of `npm ls` warning, or nothing warning, etc). By adding `optionalPeerDependencies`, all older tools would ignore it - whereas if you add a nonstandard value in the `peerDependencies` object, it will *definitely* break all current use cases.

What am I missing?

arcanis:
I'm not sure what isn't clear, so I'll do my best to answer but feel free to detail more if it doesn't answer your questions. First, my definition of "break" as far as old package managers are concerned: running `install` yields a zero exit code and no warning, but the `require` contract isn't respected.

> whereas if you add a nonstandard value in the peerDependencies object, it will definitely break all current use cases.

No. As I mentioned before, the range in `peerDependencies` is a hint, not a strict check. **It doesn't cause the install to fail.** As such, a new protocol will print warnings on older package managers (because the hint will not match), will not break anything.

So to be crystal clear: a new protocol will cause older package managers to print warnings when they encounter a new package that use it, but the `require` contract will still be respected (because the package manager will still be able to account for the peer dependency when building the dependency tree). This is not breaking.

> By adding `optionalPeerDependencies`, all older tools would ignore it

Exactly. And because they would ignore it, these package managers **would silently generate invalid dependency trees.** The users wouldn't be aware of the breakage, and worst of all it would work in most cases - meaning that the breakages would suddenly appear after running entirely unrelated commands that would cause the hoisting to change. This is breaking.

ljharb:
The syntax may not break installs, but it will break any package that uses `npm ls` in CI - which a number of packages do, including all of airbnb's and mine.

I do see the tradeoffs here, and it'd be great to chat with you offline to figure something out.

arcanis:
Sure - ping me on Discord and we can either discuss there or setup a meeting for some other time. Just as a reminder, I'll soon put together a first tentative implementation 🙂

arcanis:
Peer dependencies are implemented in https://github.com/yarnpkg/yarn/pull/6671. Feel free to review it and let me know what you think.

Still no word from npm; they've been given plenty of time and chose not to participate. Unless they share feedback now, this will land in Yarn 1.13.

zkochan:
pnpm will also support this (https://github.com/pnpm/pnpm/issues/1486)

arcanis:
I've just had a chat with @ljharb. Gist is:

- npm less than 4 install peer dependencies, so they'll break
- `npm ls` would break

I'm not too attached to those two things if there isn't a proper alternative to move forward, but he suggested to set peer dependencies in an object duplicated from 'proper' peer dependencies (basically what @zkochan suggested in https://github.com/yarnpkg/rfcs/pull/105#issuecomment-434499016, cf `strictPeers`). This would have the advantage of solving the two issues described above with minimal casualties (older package managers would simply produce warnings).

If we do this, I think we should try to be more forward compatible and make it a `peerDependenciesMeta` field (if someone has a better name ...). Something similar to this:

```json
{
  "peerDependencies": {
    "react": "^16.0.0"
  },
  "peerDependenciesMeta": {
    "react": {
      "optional": true
    }
  }
}
```

This would have the added benefit that we would be able to add additional fields in the future. For example we could add `strict` that would elevate the warning into a fatal (thus allowing the behavior that @ljharb was advocating for in a separate thread), or a `description` field that would describe why is the peer dependencies needed, or ... etc.

iarna:
peer dependencies stopped installing in npm@3, not npm@4. At the time, it was the main thing considered a breaking change. (We didn't realize, at release, the full consequences of flattening the tree, and how many people had relied on the old deeply nested tree.)

If, as a larger ecosystem, we're going to talk about adding more dependency metadata, maybe now would be a time to consider that more generally? Like, people would like to have optional dev dependencies too. And peer dependencies… well, they're used in mutually incompatible ways and if we're going to look at fixing them, maybe something more comprehensive is called for.

Peer deps are hard because of all the different meanings they carry: They can mean "I'm a plugin to this other thing, it MUST be installed and MUST be this range" or they can mean "I can take advantage of these modules, if they're installed, but they're not required. But if they are used, they MUST be these versions." They can mean "I MUST be able to require these things" or they can mean "these things MUST be able to require _me_".

(In a perfect world, I'd like to see discussions like this happen outside of our individual issue trackers. I know I've heard this request from others before, which is part of why we made package.community. (But it's also for the cross-language aspect.) But I'm open to other venues. It would just be easier to have full-ecosystem discussions in a more general space.

iarna:
I'm with @ljharb on saying that `optional:` prefixing is not a good idea. Personally I would prefer `optionalPeerDependencies` over `peerDependenciesMeta`, only because the latter is both broad and I think too narrow.

arcanis:
> If, as a larger ecosystem, we're going to talk about adding more dependency metadata, maybe now would be a time to consider that more generally? Like, people would like to have optional dev dependencies too. And peer dependencies… well, they're used in mutually incompatible ways and if we're going to look at fixing them, maybe something more comprehensive is called for.

Amen to that - if we can find a way to solve this it would be quite beneficial indeed. That said, unless we vow to reach a consensus before the end of the year, I think the best is to find a solution that doesn't prevent future evolutions rather than a complete overhaul.

> Personally I would prefer `optionalPeerDependencies` over `peerDependenciesMeta`, only because the latter is both broad and I think too narrow.

I'm a bit concerned that it will just prove painful the next time we'll need to add some flag, since we would have to either add yet another `myFlagPeerDependencies` (with the problem that it would only work if `myFlag` is a boolean), or implement something similar to `peerDependenciesMeta` (which will be weird since we'll have to keep supporting `optionalPeerDependencies`).

ljharb:
Maybe instead of "peer" then, `dependenciesOptions`? Something like:
```js
"dependenciesOptions": {
  "react": {
    "optional": true // whether it's a peer dep or not
  },
},
```

arcanis:
Maybe. Peers and non-peers have quite different semantics though so I'm not too worried about keeping their option sets apart (it's not like `dev` or `optional` which will always have the same behavior given the same dependency type).

For example, `optional` wouldn't mean the same thing in both cases: peer dependencies `optional` would be "don't warn if it's missing" while regular dependencies `optional` would be "don't fail if it fails to pass the build step". Similarly, regular dependencies could support a `dev` flag that wouldn't make sense for peer dependencies.

iarna:
I have seen requests for optional dev deps previously. Specifically around things that might improve dev experience but are not necessary to build the package. (eg, file watchers, doc generators, etc) The work around for that has been to have dev dep on a module that optionally depends on the component, requiring this kind of shim to get that behavior.

Hmm, some other notes, `optionalDependencies`, are ALSO dependencies. Declaring something solely as an optional dep implicitly adds it as a dep. (Which is why you can't list dev deps as optional deps—doing so makes them regular deps.) If we were to do `optionalPeerDependencies` it likely should behave the same.

Another thing, it's quite normal to have something as both a peer dep AND a dev dep. Which raises a concern about `dependenciesOptions` being somewhat confusing as to what it should do when matching both. (Like, you can legitimately want a non-optional dev dep, but an optional peer dep.)

I'm curious what other kinds of options you imagine being added specifically to peer dependencies in the future?

arcanis:
> I'm curious what other kinds of options you imagine being added specifically to peer dependencies in the future?

For peer dependencies in particular I'd say `required` (missing is a hard failure), `optional` (missing is silent), `strict` (mismatching range is a failure - not 100% convinced about this one). The `description` one could be useful when used together with `optional`, since it would allow to suggest packages to the user (similar to what `apt-get` does). Right now I don't see immediate others, but I'm certainly missing some.

For meta in general I'm not sure, but I think it's safe to say that there will be opportunities. Documenting dependencies has been a longstanding request for example (which prompted the support of the `//` key, which is impractical for various reasons), and that could be solved with a meta field (similar to the `description` one listed above). A `condition` meta field similar to what Media Queries are could also be useful to only add dependencies under certain conditions (like the os, or the project config, ...), etc.

A quite different use case, but meta properties could be used to configure particular resolver behaviors. For example to specify a preferred registry for resolution, etc.

> I have seen requests for optional dev deps previously. Specifically around things that might improve dev experience but are not necessary to build the package.

I'm curious - do you know what are the most common packages listed as `optionalDependencies`? The only one I consistently see is `fsevents`, I wonder if it would make sense to fix it altogether so that it would always build to *something*. Allowed-to-fail dependencies in general seem weird to me since they are generally just as easy to implement in userland (`"postinstall": "doesnt-compile || true"`).

arcanis:
https://github.com/yarnpkg/yarn/pull/6671 has been updated with `peerDependenciesMeta` (and support for `optional:` has been reverted).

iarna:
What do you think about making peer dependencies optional-by-default, only complaining if there's an incompatible peer dep installed?

And instead have some facility for making them required.

I ask this, because as things stand today, I'm not sure peer dependency warnings provide any value… like, yes, installing a plugin without the base module is an error, it's also gives a pretty straightforward error when you try to use it (in fact, most plugins there's no way to consume it without the base module anyway).

ljharb:
That contradicts the way most of the ecosystem uses them - in eslint, react, enzyme, babel, and webpack’s ecosystems, at least, a peer dep warning means nothing can be expected to work because the dep graph is invalid.

It’s not about a missing base as much as it’s about having the wrong version - which can fail in subtle ways. Peer deps are the way that’s been communicated in my experience.

iarna:
@ljharb I'm suggesting we would still warn on wrong version. Just not on "missing entirely", which tends to happen with things like `discord.js` that have a bunch of peer deps on modules that it _can_ use, optionally, but does not require.

arcanis:
I don't have super strong feelings regarding whether the default should be "optional" or not 🤔 

The cautious side in me would probably slightly go towards moving behind an option in order to preserve the existing behavior, but both are acceptable to me.

arcanis:
It seems we have a reasonable consensus, so I'll merge the PR tomorrow unless you have objections 🙂

@iarna Does npm have an Intent to Support, like pnpm?

ljharb:
Given that today is a US holiday and many people are traveling, perhaps it would be prudent to wait until Monday, so more folks have the chance to weigh in?

arcanis:
Oh indeed - sorry, good holidays everyone! 🎉

arcanis:
`peerDependenciesMeta` has been merged into master (https://github.com/yarnpkg/yarn/commit/011a634d787611e9c1f2f1b888212525f1fc096d)

arcanis:
Since I think we've reached a consensus, so I've merged the RFC (I forgot to update it to reflect the changes we've discussed, so I've committed it separately: 835c3278d8afb7665b841939fc22d7c7433cb868). Will be part of the soon-to-be-released 1.14! 🙂

arcanis:
I'm not too convinced by this use case, but maybe. Feel free to write a formal rfc to list pros and cons 🙂

zkochan:
I wonder if we should support a reacher syntax. A package may support hundreds of plugins. For pnpm, it would be important to know about these relations. For instance.

```json
"peerDependencies": {
  "foo-plugin-*": "*"
},
"metaPeerDependencies": {
  "foo-plugin-*": {
    "optional": true
  }
}
```

arcanis:
Hm I understand the benefits (would be useful for things like `babel` and its plugins), but at the same time I think it'd be for the best if we could keep the system simple.

Wouldn't this use case be solved through userland btw? Something like this:

**app/index.js**
```js
const eslint = require('eslint');
eslint.findPlugins(__dirname);
```

**eslint/index.js**
```js
exports.findPlugins = p => {
    const pkgJson = require(path.join(p, `package.json`));
    const peerDependencies = Object.keys(pkgJson.dependencies).filter(name => {
        return name.test(/eslint-plugin-*/);
    }).map(name => {
        return require(name, {paths: [p]});
    });
};
```

ljharb:
Using any form of "one to many" in dependency lists seems like it would vastly inflate the complexity of determining a dependency graph, including that the packages eligible to be included in the glob could change at any time based on network activity (as opposed to the current situation, where basically only versions could change at any time)

zkochan:
> Using any form of "one to many" in dependency lists seems like it would vastly inflate the complexity of determining a dependency graph, including that the packages eligible to be included in the glob could change at any time based on network activity

That's because Yarn/npm create a flat node_modules, right? That randomness cannot happen with pnpm's nested node_modules.

I like the solution proposed by @arcanis.

ljharb:
No, it has nothing to do with node_modules on disk - i mean that you couldn’t look at a package.json and get a list of the top-level dependencies needed. There’s no randomness about yarn/npm that doesn’t exist with pnpm and everything else, when someone newly publishes a package that matches an existing glob.

PR #111, opened 2018-12-06
Add "410 Gone" RFC for unpublished modules

markstos:
Unpublished packages currently cause pain for yarn users. A recommended approach ends up being to [delete yarn.lock and re-generate it](https://github.com/yarnpkg/yarn/issues/6702#issuecomment-442314783), which can cause a huge amount of dependencies to change when only one was needed.

Unpublished packages currently return the HTTP status code "404 Not Found". This code is designed for resources that might exist again in the future. In the NPM ecosystem, version numbers are wisely [immutable](https://www.npmjs.com/policies/unpublish). A better HTTP status code to return for an unpublished package is "410 Gone", designed for resources that will never come back.

When `yarn` encounters a [410 Gone](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/410) resource, it could try to resolve the related semantic version again, possibly selecting a newer version, fixing the issue the single dependency without requiring regenerating all of "yarn.lock".

### DIFF
--- a/accepted/0000-410-gone-for-unpublished.md
+++ b/accepted/0000-410-gone-for-unpublished.md
@@ -1,0 +1,96 @@
+- Start Date: 2018-12-06 (fill me in with today's date, YYYY-MM-DD)
+- RFC PR:
+- Yarn Issue:
+
+# Summary
+
+Unpublished packages currently cause pain for yarn users. A recommended approach
+ends up being to [delete yarn.lock and re-generate it](https://github.com/yarnpkg/yarn/issues/6702#issuecomment-442314783),
+which can cause a huge amount of dependencies to change when only one was needed.
+
+Unpublished packages currently return the HTTP status code "404 Not Found".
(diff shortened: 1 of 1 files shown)


ljharb:
Status codes come from the public registry, which has nothing to do with yarn itself. Is this a chance the npm registry is planning to make?

markstos:
@ljharb Yarn chooses to setup `registry.yarnpkg.org` as the source for most package downloads, not `npmjs.org`. This proxy can choose to return it's own codes, however, I'm not sure this proxy is aware of which modules have been unpublished. I have not been in conversation with npm, Inc about this proposal thus far.

ljharb:
I'd suggest starting a conversation with npm - since as you say, even if the CNAME was set up as an actual server, it'd still have no way to know what's been unpublished.

markstos:
@ljharb Thanks for the feedback. I'll do that and report back.

PR #112, opened 2019-01-07
[RFC] Yarn CI

thynson:
This RFC aims to speeding up CI workflow by introduce a new command `yarn ci`. It shall be better than `npm ci`.

### DIFF
--- a/accepted/0000-yarn-add-from-only-lockfile.md
+++ b/accepted/0000-yarn-add-from-only-lockfile.md
@@ -1,0 +1,70 @@
+- Start Date: 2018-12-12
+- RFC PR:
+- Yarn Issue:
+
+# Summary
+
+This RFC relaxed the requirement to install package into local `node_modules`,
+to make it more friendly to layered file system that used by docker and
+common CI cache system.
+
+# Motivation
+
+NPM projects model requires all project to specify their dependencies as well
+as project metadata, e.g. name, version, scripts, in one file called
+`package.json`. However this is not cache-friendly with layered file system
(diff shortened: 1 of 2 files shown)


bestander:
1. A lock file can contain multiple versions of the same package, without package.json there would be no way to know which one of them is to be hoisted at the top.

2. Name 'ci' is confusing, clashes with npm's command and usually yarn names commands as simple verbs.

3. The use case looks very niche. 
- With offline mirror you don't need to be worried about clearing cache from internet connection point of view
- With pnp you don't need to worry about performance
- The problem is only true when you change package.json file with no dependency changes. I think it's easier to avoid change to package.json at application level than introduce a new command to yarn

bestander:
I'll close but feel free to make arguments

ljharb:
`npm ci` is short for `npm clean-install`; the use case is, unlike a normal install, which reads node_modules on disk, and updates it to contain the needed packages; ci rimrafs node_modules and then installs from scratch. It may be that yarn install can’t get any speedup from this approach; npm does. I don’t think it has anything to do with package.json whatsoever.

arcanis:
In general I'm not too fond of making installs faster by implementing a second command. I would very much prefer default installs to be fast (or decrease the amount of installs required), otherwise only a fraction of the use cases will be improved.

I think in our case this is done by PnP, which sensibly improves the install time. The v2 will also offer the ability to check-in a clean set of "install artifacts" in the repository, meaning you won't even need to run `yarn install` in most cases.

PR #113, opened 2019-01-18
Enhance `yarn <argv>` command to lookup global modules registry

thymikee:
Inspired by [`npx`](https://medium.com/@maybekatz/introducing-npx-an-npm-package-runner-55f7d4bd282b), enhance `yarn <argv>` command to lookup global modules registry and prompt to install if there's no such module or it's outdated.

### Example:

When there's no `react-native` global binary:

```sh
yarn react-native init MyApp
info: binary not found; use "react-native@0.59" from registry? (Y/n)
# fetching react-native, populating yarn global cache
# running global react-native init MyApp
yarn react-native run-ios # from local install
```

Or

```sh
# `node` is not in package.json
yarn node@6 compat-script.js
```

cc @arcanis @cpojer @rickhanlonii

### DIFF
--- a/accepted/0000-yarn-global-execute.md
+++ b/accepted/0000-yarn-global-execute.md
@@ -1,0 +1,80 @@
+- Start Date: (fill me in with today's date, YYYY-MM-DD)
+- RFC PR: (leave this empty)
+- Yarn Issue: (leave this empty)
+
+# Summary
+
+Inspired by [`npx`](https://medium.com/@maybekatz/introducing-npx-an-npm-package-runner-55f7d4bd282b), enhance `yarn <argv>` command to lookup global modules registry and prompt to install if there's no such module or it's outdated.
+
+# Motivation
+
+There are packages (CLIs) that both serve as global and local binaries, examples are `react-native`, `jest`, `eslint`, `ember-cli` to name a few.
+With `npx` we can do:
+
+```sh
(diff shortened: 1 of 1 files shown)


arcanis:
I have concerns about this syntax change:

- The `yarn <script>` syntax currently only executes local scripts. It would be confusing imo to make it execute global scripts based on an hidden state (whether or not a package exists or not).

- It's not clear what happens in the case where you're on a freshly cloned project and run `yarn react-native`. In this situation Yarn doesn't know yet what are the local binaries, meaning that you must either 1/ require an install if the command doesn't exist (confusing), or 2/ install `react-native` globally and run it (dangerous), 3/ ask for the user to run `yarn install` (I could support this one, if it wasn't for the next point). 

- Running `yarn <script>` is guaranteed to either execute the script (if it exists), or fail (if it doesn't). By adding an interactive prompt we effectively disable this behavior which might have far-reaching effects for scripts. It could be scoped to TTYs only, but I would still be wary.

- It opens the way for things to work locally and break on different environments (`yarn eslint` would work even if you forgot to list `eslint` in your dependencies, but would later break on your CI), which is something we've been fighting against until now.

- It could be opt-in rather than opt-out, but realistically noone would think about enabling it.

thymikee:
Closing in favor of https://github.com/yarnpkg/berry/pull/40

PR #116, opened 2019-05-17
Add importable scripts

cvanem:
Allow scripts to be imported from a file as an alternative to the standard object hash in the packages.json scripts field. Also define the basic structure of the imported file. 

[RFC](https://github.com/yarnpkg/rfcs/pull/116/files?short_path=dbe3889#diff-dbe3889482537688c3ecd635b8cc8a4e)
Documentation: https://github.com/yarnpkg/website/pull/951
PR: https://github.com/yarnpkg/yarn/pull/7284

### DIFF
--- a/accepted/0000-importable-scripts.md
+++ b/accepted/0000-importable-scripts.md
@@ -103,12 +103,10 @@ This feature can be introduced via documentation as an optional alternative to i
 # Drawbacks
 
 1.  Increase binary size.  However small, there will be a slight increase in the binary size.
-2.  There will probably be some additional overhead for people having issues or understanding 
-    how to implement the more advanced structures.
+2.  There will probably be some additional overhead for people having issues or understanding how to implement the more advanced structures.
 3.  More time needed to create documentation.
-4.  As the package.json script field will now be used to specify a filename, we need to determine
-    if this will have any effect on other libraries or tools.  One thing I noted is that VS Code
-    does highlight this field with an 'Incorrect type. Expected Object' warning when a filename is specified.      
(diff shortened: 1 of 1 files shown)


arcanis:
Hey! Thanks for the suggestion 😃 I do have one big concern, however: this proposal makes it impossible to statically know what scripts are available. We would always have to execute the Javascript code to figure it out.

cvanem:
@arcanis Yes, I suppose that is another drawback that should be added.  Is this concern due to a performance hit or something else?  The Javascript code would only be executed if a filename input is provided.

cvanem:
@arcanis Hey just following up on this.  Can you tell me how likely it is that this [PR](https://github.com/yarnpkg/yarn/pull/7284) will be accepted anytime soon?  The reason I am asking, is that we are currently evaluating some options for cleaning up our scripts.  See https://github.com/mui-org/material-ui/issues/15497

If this is not going to be accepted, then I would like to resolve any issues so it can be accepted, or look at other options, such as using [nps](https://github.com/kentcdodds/nps).

zkochan:
This is one of the reasons I added support of `package.yaml` to [pnpm](https://github.com/pnpm/pnpm) (although I am still not sure that was a good idea). But a yaml file (or a json5 file), also satisfies your requirements.

You can write scripts like this:

```yaml
scripts:
  # This is some comment
  test: >-
    pnpm recursive test
    --reporter append-only
    --workspace-concurrency 1
    --filter ./packages
    --filter ./privatePackages
build: pnpm recursive run tsc
```

PR #120, opened 2020-08-07
Preload `dependenciesMeta` for packages

ylemkimon:
[Rendered](https://github.com/yarnpkg/rfcs/blob/0112277b8c19af2ede79952edaa67be14c58f78e/accepted/0000-preload-dependencies-meta.md)

### DIFF
--- a/accepted/0000-preload-dependencies-meta.md
+++ b/accepted/0000-preload-dependencies-meta.md
@@ -1,0 +1,84 @@
+- Start Date: 2020-08-07
+- RFC PR: (leave this empty)
+- Yarn Issue: (leave this empty)
+
+# Summary
+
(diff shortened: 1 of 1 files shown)


arcanis:
It's already the case! The compat plugin contains various builtin extensions - we just need to fill it more aggressively, but since we require to also send a PR upstream, it takes a bit of time that core maintainers don't always have:

https://github.com/yarnpkg/berry/blob/master/packages/plugin-compat/sources/extensions.ts

arcanis:
Ah nevermind, I misunderstood your PR, you're talking about security, not compatibility. I'll close this RFC and reopen the issue to discuss it, as we don't really use this repository anymore.

PR #121, opened 2020-08-16
Added RFC for "yarn show"

bhushan-shah-dev:
Added RFC for a new command `yarn show`, inspired by [this issue](https://github.com/yarnpkg/yarn/issues/3569).

### DIFF
--- a/accepted/0000-yarn-show.md
+++ b/accepted/0000-yarn-show.md
@@ -1,0 +1,51 @@
+- Start Date: (fill me in with today's date, 2020-08-16)
+- RFC PR: (leave this empty)
+- Yarn Issue: (leave this empty)
+
+# Summary
+
+Add a new command `show`, which would list the dependencies explicitly specified in `package.json`.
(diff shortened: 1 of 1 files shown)


ljharb:
The spelling “show” conflicts conceptually with `npm show`; I’d suggest finding an unused name.

bhushan-shah-dev:
Good point.

How about `yarn explicit`?

fivethreeo:
Explicit is a «keyboardful» yarn showdeps or yarn versions or yarn dependencies

kesor:
If someone is actually using `yarn list --depth=0` and parses the output, and after that issue is fixed they would get a different output ... is it *really* a "breaking" change?

Is it really so essential to introduce yet another subcommand that does exactly what 99% of people intuition is already telling them that `yarn list --depth=0` should be doing?

Introducing yet another command is bad UX. Good UX is when you do what the people feel intuitively should "just work".

